PR #5, opened 2023-01-12
Assignment1 with @mhtran23

BenagainB:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -54,7 +54,7 @@ public static bool DisplayResult(string userGuess, Question question)
 
         public static void DisplayQuestion(Question question)
         {
-            Console.WriteLine("Question: " + question.Text);
+            Console.WriteLine("Question: " + question.Text); //System.NullReferenceException thrown here
```
crawfordkid2, line 57: What is null here? specify on your comments to allow clarity for yourself and others. formatting wise I would put the comment on the line above not on the same line.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,6 +89,7 @@ public static Question[] LoadQuestions(string filePath)
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question; //added to resolve NullReferenceException
```
crawfordkid2, line 92: in the comment describe how it solves the issue, good documentation to show where is changed but seems like a better comment to leave for the update to your repo when you stage the changes on git.

crawfordkid2 (commented):
Good solution, format comments differently if you see fit, there is more useful things to comment such as specifics or what you are doing when you write the new code. However, comments are also not needed for well written code, maybe save your comments for when you stage the changes on git.

mhtran23:
@JessicaH46 @crawfordkid2 unit test for the percentage is fixed. thank you

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
BenagainB: Thanks for the feedback. I removed inline comments since that info should be in the change summary.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
BenagainB: Thanks. I followed your advice.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,22 +22,47 @@ public static void Main(string[] args)
...
+            
+            
             return DisplayResult(userGuess, question);
         }
 
-        public static string GetGuessFromUser()
+        public static string GetGuessFromUser(Question question)
         {
-            return Console.ReadLine();
+            bool indexInBounds = true;
```
mmwoodfo, line 42: Wouldn't you want to name this variable `indexOutOfBounds`?

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,22 +22,47 @@ public static void Main(string[] args)
...
+            string guess = Console.ReadLine();
+            while (indexInBounds)
+            {
+
+                if (guess == "1" || guess == "2" || guess == "3")
+                {
+                    indexInBounds = false;
+                }
+                else if (guess is "x" or "X")
```
mmwoodfo, line 51: You can also use the built-in method `.ToLower()` here, so you don't have to repeat your check for if `x` was typed.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,22 +22,47 @@ public static void Main(string[] args)
...
-        public static string GetGuessFromUser()
+        public static string GetGuessFromUser(Question question)
         {
-            return Console.ReadLine();
+            bool indexInBounds = true;
+            string guess = Console.ReadLine();
+            while (indexInBounds)
+            {
+
+                if (guess == "1" || guess == "2" || guess == "3")
```
mmwoodfo, line 47: Another option, would be to convert the string to a number with [`int.TryParse`,](https://learn.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-7.0) use `<` or `>` and loop if the number conversion fails (since characters are not allowed)

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,22 +22,47 @@ public static void Main(string[] args)
...
+                {
+                    indexInBounds = false;
+                }
+                else if (guess is "x" or "X")
+                {
+                    Environment.Exit(10);
+                }
+                else
+                {
```
mmwoodfo, line 56: Good idea for the extra credit. :)

PR #9, opened 2023-01-14
Assignment1 - Initial Submit

FenrisStargazer:

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -72,6 +72,21 @@ namespace PrincessBrideTrivia.Tests
             Assert.AreEqual(expectedString, percentage);
         }
 
+        [TestMethod]
+        public void GetHintTxt_ProperlyChangesHintTxt()
+        {
+            string v1 = "you didn't use any hints!";
+            string v2 = "you only used 1 hint!";
+            string v3 = "you only used 5 hints!";
+            string t1 = Program.GetHintTxt(0);
+            string t2 = Program.GetHintTxt(1);
+            string t3 = Program.GetHintTxt(5);
+            Assert.AreEqual(t1, v1);
+            Assert.AreEqual(t2, v2);
+            Assert.AreEqual(t3, v3);
+
+        }
+
         private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
         {
             for (int i = 0; i < numberOfQuestions; i++)
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -72,7 +72,6 @@ namespace PrincessBrideTrivia.Tests
             Assert.AreEqual(expectedString, percentage);
         }
 
-
         private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
         {
             for (int i = 0; i < numberOfQuestions; i++)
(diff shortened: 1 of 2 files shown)


shweshipu:

    Issue 3: Feature request unit tested ❌

I'm not really sure how you could unit test the colors. The hints feature could probably be unit tested though.

Nice work 👍

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -5,31 +5,57 @@ namespace PrincessBrideTrivia
...
                 {
                     numberCorrect++;
-                }
+                } 
             }
-            Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
+            string hintTxt;
+            if (Hints == 0)
+            {
+                hintTxt = "you didn't use any hints!";
```
DaisyPalencia, line 29: I would recommend deciding whether to have the starts of your strings be capital or lowercase and sticking with it. It would make it cleaner, because right now some start with "You" but then others start as "you".

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -5,31 +5,57 @@ namespace PrincessBrideTrivia
...
+            } else
+            {
+                hintTxt = "you only used " + Hints + " hints!";
+            }
+            Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct and " + hintTxt);
+        }
+
+        public static string GetPercentCorrect(double numberCorrectAnswers, int numberOfQuestions)
+        {
+            double num1 = numberCorrectAnswers;
```
DaisyPalencia, line 42: This works but it can be simplified to make your code easier to read. Instead of saving the numberCorrectAnswers and numberOfQuestions as 2 new variables you can just "return ((double)numberCorrectAnswers / (double) numberOfQuestions * 100) + "%"".

DaisyPalencia (commented):
Come up with a new feature for the application
Issue 3: Feature request unit tested ❌

Good job! Your feature was really cool.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
FenrisStargazer: I did the ones related to the hint text as lowercase because on line 37 I concatenate that with the initial part that says "You got x% correct" so it prints as "You got x% correct and only used x hints!". The if-block just determines if they used 0, 1 or more hints so the wording ends up cleaner (e.g. you don't have "you used only 1 hints" or "you used only 0 hints"). Though if there are inconsistencies outside of that section then I definitely just missed something.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
FenrisStargazer: Yeah I'm definitely going to be changing it. For some reason when I first tried casting them as doubles in the beginning it wasn't working, I'm not even sure what I did different.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -42,24 +74,54 @@ public static string GetGuessFromUser()
 
         public static bool DisplayResult(string userGuess, Question question)
         {
+            if (userGuess == "?")
+            {
+                _Hints++;
```
mmwoodfo, line 79: If I try to use a hint twice in the same question, this counts as part of my total hint count at the end.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +72,20 @@ public void GetFilePath_ReturnsFileThatExists()
             Assert.AreEqual(expectedString, percentage);
         }
 
+        [TestMethod]
+        public void GetHintTxt_ProperlyChangesHintTxt()
```
mmwoodfo, line 76: Another test for making sure that only 1 hint per question is printed out would be good to have as well!

mmwoodfo (commented):
Good job! Nice extra credit idea!

PR #14, opened 2023-01-15
Assignment 1 - Zach Lawson

zlawson-ewu:
Added casts to GetPercentCorrect division to fix failing tests. 
Added assignment of newly created Question object to question array in LoadQuestions. 
Added feature to rank player performance with a string message based on percent correct. 
Added test for new feature.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
         }
 
         public static bool AskQuestion(Question question)
@@ -89,6 +89,8 @@ namespace PrincessBrideTrivia
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+
+                questions[i] = question;
             }
             return questions;
         }
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -19,12 +19,32 @@ namespace PrincessBrideTrivia
                     numberCorrect++;
                 }
             }
-            Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
+            string percentCorrect = GetPercentCorrect(numberCorrect, questions.Length);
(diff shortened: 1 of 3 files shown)


Tom-Rohr:
I will review this

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -19,12 +19,32 @@ public static void Main(string[] args)
...
-            Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
+            string percentCorrect = GetPercentCorrect(numberCorrect, questions.Length);
+            Console.WriteLine("You got " + percentCorrect + " correct");
+            Console.WriteLine(PercentCorrectRatingMessage(percentCorrect));
         }
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (int)((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
Tom-Rohr, line 29: You might consider rounding to the nearest hundredth of a percent with Math.round() method, just for aesthetic purposes when the score is printed.

Tom-Rohr (commented):
Nice work. I like the feature you added of printing different messages based on the user's end score.

* Pull Request title includes "Assignment 1" in the title ✔❌
* Pull Request targets Assignment 1 branch ✔❌
* Issue 1: Application no longer crashes ✔❌
* Issue 2: Unit test properly passes ✔❌

eroadhouse:
I'll do the second review here

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
eroadhouse, line 29: I think this is a good way of handling the problem with the double division, casting it to a double and then casting that final product to an int. Nice work

eroadhouse (commented):
Well done, all tests pass, no crashes or any incorrect output, and the new feature is clever!

Issue 3: Feature request unit tested ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -19,12 +19,32 @@ public static void Main(string[] args)
...
+            }
+            if (rank >= 70)
+            {
+                return "Good enough, but can you really call yourself a 'true fan' with a score like that?";
+            }
+            if (rank >= 40)
+            {
+                return "You've surely seen this movie before, but how long ago was that?";
+            }
+            return "Have you even seen this movie? Try again after watching!";
```
JLee35, line 47: Nice new feature. A good practice is to have only one return statement in a code block (if possible). A suggestion here would be to store the result in a variable and have only one return statement at the end. This would be a good place for a switch statement.

JLee35:
Good job. Remember to at least acknowledge peer reviews and resolve conversations once the comment has been addressed.

PR #18, opened 2023-01-16
Assignment1

shweshipu:
Issues 1 and 2 were solved

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -89,6 +89,8 @@ namespace PrincessBrideTrivia
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+
+                questions[i] = question;
             }
             return questions;
         }
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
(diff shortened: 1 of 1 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
mmoran314, line 27: Pull Request title includes "Assignment 1" in the title ✔
I'm not sure if it is something worth changing, but changing the arguments in the GetPercentCorrect method to doubles will take away the need to use casting on the variables in the percent correct calculation.
Other than that, it looks great and I do not see anything that would need to be changed.

mmoran314 (changes requested):
Issue 2: Unit test properly passes  ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
Joshua-Pedersen, line 27: You could simplify this by just casting once. A ```double``` divided by an ```int``` will result in a ```double```
~~~
return ((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
~~~

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,6 +89,8 @@ public static Question[] LoadQuestions(string filePath)
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+
+                questions[i] = question;
```
Joshua-Pedersen, line 93: Not sure if they are requiring this, but the added code isn't covered by a test. Could create a test to confirm the ```question``` is being placed into the ```questions``` array

Joshua-Pedersen (commented):

Changes clear and working.

PR #19, opened 2023-01-16
Assignment1

Tom-Rohr:
Sorry, no partner for this assignment. I will have one for all remaining assignments.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -86,5 +86,18 @@ namespace PrincessBrideTrivia.Tests
                 File.AppendAllLines(filePath, lines);
             }
         }
+
+        [TestMethod]
+        public void OfferCorrectAnswer_ReturnsCorrectAnswer()
+        {
+            Question testQuestion = new Question();
+            testQuestion.Text = "What color is the sky?";
+            testQuestion.Answers = new string[] {"Red", "Orange", "Blue" };
+            testQuestion.CorrectAnswerIndex = "2";
+
+            string correctAnswer = Program.GetCorrectAnswer(testQuestion);
+
+            Assert.AreEqual(correctAnswer, testQuestion.Answers[int.Parse(testQuestion.CorrectAnswerIndex)-1]);
+        }
     }
 }
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -10,7 +10,7 @@ namespace PrincessBrideTrivia
             string filePath = GetFilePath();
             Question[] questions = LoadQuestions(filePath);
 
-            int numberCorrect = 0;
+            double numberCorrect = 0;
             for (int i = 0; i < questions.Length; i++)
             {
                 bool result = AskQuestion(questions[i]);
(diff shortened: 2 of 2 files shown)


zlawson-ewu:
I will review this.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -10,21 +10,25 @@ public static void Main(string[] args)
...
+                else
+                {
+                    OfferCorrectAnswer(questions[i]);
+                }
             }
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
-        public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
+        public static string GetPercentCorrect(double numberCorrectAnswers, int numberOfQuestions)
```
zlawson-ewu, line 29: Functions correctly. Complexity could be reduced with the use of casts within GetPercentCorrect, but this solution works fine.

zlawson-ewu (approved):
Overall, good job! The program functions as-expected, the new feature works as-expected, all tests pass.

    Issue 3: Feature request unit tested ✔

Afterlyte:
Will review

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -52,6 +56,32 @@ public static bool DisplayResult(string userGuess, Question question)
...
 
+        public static void OfferCorrectAnswer(Question question)
+        {
+            string wantsAnswer = "";
+
+            Console.WriteLine("Would you like to see the correct answer? Type Y or N");
+            wantsAnswer = Console.ReadLine();
+            wantsAnswer = wantsAnswer.ToLower();
+
+            while (!wantsAnswer.Equals("y") && !wantsAnswer.Equals("n"))
```
Afterlyte, line 67: It's probably a good idea to print a line that specifies why you're prompting again, even though it's pretty obvious in this case it's still good for ease of use and understanding.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -52,6 +56,32 @@ public static bool DisplayResult(string userGuess, Question question)
             return false;
         }
 
+        public static void OfferCorrectAnswer(Question question)
```
Afterlyte, line 59: I really like this idea overall, it's good for the user whether they want to do a one-and-done run of the quiz or keep trying until they get them all right.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
Afterlyte, line 29: While this does work, it might read a little weird to others, especially since only _one_ of them is set to a double, also since the number correct can never be a decimal number.

Another way of doing it could be to tinker with the original equation a little (check the position of the "* 100"), so that the integer division will still produce the correct result _and_ will round for you, or to use casts when doing the calculations so it's clear you're just using double for the calculation portion.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +86,18 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
...
+        public void OfferCorrectAnswer_ReturnsCorrectAnswer()
+        {
+            Question testQuestion = new Question();
+            testQuestion.Text = "What color is the sky?";
+            testQuestion.Answers = new string[] {"Red", "Orange", "Blue" };
+            testQuestion.CorrectAnswerIndex = "2";
+
+            string correctAnswer = Program.GetCorrectAnswer(testQuestion);
+
+            Assert.AreEqual(correctAnswer, testQuestion.Answers[int.Parse(testQuestion.CorrectAnswerIndex)-1]);
```
Afterlyte, line 100: I like this approach of showing that the correct answer can be fetched for offering the answer, but you might want to tweak this to show that the OfferCorrectAnswer method you wrote is necessary. You wrote the method in the program but this test never utilizes it to show that it needed to be there.

Afterlyte (commented):

mmwoodfo (commented):
No problem on the partner, I'll give you a pass for this assignment. Thank you for adding the note about it.

In the future, be sure to address your PR comments, there are a number of suggestions that have been left unacknowledged.

PR #20, opened 2023-01-16
Assignment 1

Malcolm-Shepherd:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (int)((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
DaisyPalencia, line 27: For this you don't need to cast the line as an int, so I would recommend to remove the int from your code.

Wvvern (commented):
- Pull Request title includes "Assignment 1" in the title ✔
- Pull Request targets Assignment 1 branch ✔
- Issue 1: Application no longer crashes ✔
- Issue 2: Unit test properly passes ✔

DaisyPalencia (commented):

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
JLee35: The 'int' cast actually makes the result cleaner for the user ('85%' versus '85.67547%'), so I think it is a great addition.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
JLee35, line 27: Casting to an `int` or using `math.round` would have cleaned this up for the user.

JLee35:
Ensure to address review comments next time.

PR #21, opened 2023-01-16
Assignment1

randypipkins:
Program.cs changed so the unit test passes and program runs


jstanley6:
I will review this

jstanley6 (commented):

jstanley6:

jstanley6:
Only thing I would do to this maybe is add a few error checking to double check the file isn't empty so the program doesn't crash but the program does what it is supposed too.

PR #26, opened 2023-01-17
Assignment1

jstanley6:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,9 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            double numCorrectAns = numberCorrectAnswers;
```
mmoran314, line 27: Pull Request title includes "Assignment 1" in the title ✔
Looks good!
Only potential change I can think of is to either cast the arguments as doubles during the calculation or to change the arguments within the header of the GetPercentCorrect method from ints to doubles in order to not use as many lines of code.

mmoran314 (changes requested):

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
jstanley6: Completed your suggestions.

Wvvern:
Will review

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,9 +22,9 @@ public static void Main(string[] args)
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
-        public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
+        public static string GetPercentCorrect(double numberCorrectAnswers, double numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((numberCorrectAnswers / numberOfQuestions) * 100.0) + "%";
```
Wvvern, line 27: Since both numberCorrectAnswers and numberOfQuestions are stored as ints, the method parameters here should reflect that being the case. It's also not necessary to change them to doubles.

```suggestion
           double percentage = 100.0 * numberCorrectAnswers / numberOfQuestions;
           return Math.Round(percentage, 2) + "%";
```

Instead, something like this will prevent integer division from occuring. 
You could just explicitly cast to a double here, but even that is not necessary. 
Multiplication and division are of equal weight within the order of operations, so the multiplication will occur first. 
Once the multiplication occurs, the numerator becomes a double and forces the denominator to be promoted to a double. Thus, resolving the integer division problem.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -68,6 +72,10 @@ public static string GetFilePath()
 
         public static Question[] LoadQuestions(string filePath)
         {
+            if(filePath == null)
```
Wvvern, line 75: This check is not necessary. In c#, a null value should never be passed to a non-nullable parameter, and is treated as a warning. The parameter "filePath" has no question mark at the end, indicating (by default) that it's non-nullable.
Instead, you could check if it's a valid file path.

Wvvern (commented):

PR #33, opened 2023-01-17
Assignment 1 - Tricia Faust

TriciaFaust:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -0,0 +1,106 @@
...
+                Question question = new Question();
+                question.Text = questionText;
+                question.Answers = new string[3];
+                question.Answers[0] = answer1;
+                question.Answers[1] = answer2;
+                question.Answers[2] = answer3;
+                question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question;
+            }
+            return questions;
```
randypipkins, line 103: Perfect solution! Only suggestion would be to add a check on the array that the questions are being loaded in to to ensure that it is not null or empty to avoid potential crashes along

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -0,0 +1,106 @@
...
+                        result = AskQuestion(questions[i]);
+                    }
+                }
+            }
+            Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
+        }
+
+        public static string GetPercentCorrect(double numberCorrectAnswers, double numberOfQuestions)
+        {
+            string percentage = ((numberCorrectAnswers / numberOfQuestions) * 100) + "%";
```
Malcolm-Shepherd, line 35: You should just be able to return this in one line if you wanted, but it's functionally the same.

Malcolm-Shepherd:
## Assignment Details

* Pull Request title includes "Assignment 1" in the title ❌
* Pull Request targets Assignment 1 branch ✔
* Issue 1: Application no longer crashes ✔
* Issue 2: Unit test properly passes ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
TriciaFaust: Good point, thanks

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
TriciaFaust: That's a good suggestion, saves on lines of code.  Thanks.

JLee35:
Unfortunately, your branch is missing critical files. There isn't a solution file (`.sln`) or a test file. Not sure if these got accidently deleted, but I cannot compile or run your code without the solution file.

PR #35, opened 2023-01-17
Assignment1

j0shh3ss:
Brendan Hopkins

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -11,12 +11,34 @@ namespace PrincessBrideTrivia
             Question[] questions = LoadQuestions(filePath);
 
             int numberCorrect = 0;
+            //For loop for getting whether or not the user answers correctly
             for (int i = 0; i < questions.Length; i++)
             {
                 bool result = AskQuestion(questions[i]);
+                //If correct
                 if (result)
                 {
                     numberCorrect++;
+                }
+                //If not correct
+                if (!result)
+                {
+                    Console.WriteLine("Would you like to try again? Y/N");
+                    string userGuess = GetGuessFromUser();
+
+                    if (userGuess == "y" || userGuess == "Y")
+                    {
+                        AskQuestion(questions[i]);
+                        if (result)
+                        {
+                            numberCorrect++;
+                        }
+                        //Resumes to the next question if the user replies n or N
+                        if (userGuess == "n" || userGuess == "N")
+                        {
+                            AskQuestion(questions[i++]);
+                        }
+                    }
                 }
             }
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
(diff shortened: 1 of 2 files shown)


jstanley6 (commented):
I will review this,

VeraKorchemnaya (commented):
Pull Request title includes "Assignment 1" in the title ❌ (needs a space between Assignment and 1)

Extra Credit
Issue 3: Feature request unit tested ❌ (can you think of someway to test your new feature?)

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -11,20 +11,42 @@ public static void Main(string[] args)
...
+                    Console.WriteLine("Would you like to try again? Y/N");
+                    string userGuess = GetGuessFromUser();
+
+                    if (userGuess == "y" || userGuess == "Y")
+                    {
+                        AskQuestion(questions[i]);
+                        if (result)
+                        {
+                            numberCorrect++;
+                        }
```
JLee35, line 35: `result` on line 32 will never evaluate to true because you aren't reassigning it the result from `AskQuestion` on line 31. So even if the user gets the answer right on the second try, the their overall score will not improve.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -11,20 +11,42 @@ public static void Main(string[] args)
...
+                        //Resumes to the next question if the user replies n or N
+                        if (userGuess == "n" || userGuess == "N")
+                        {
+                            AskQuestion(questions[i++]);
+                        }
```
JLee35, line 40: This will also never evaluate to true because you aren't getting the guess from the user again after it has evaluated to "y" or "Y" on line 29.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -11,20 +11,42 @@ public static void Main(string[] args)
...
+                    }
+                }
             }
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
JLee35, line 49: You only need to cast one of these to a `double`, but that is not a big deal. Also, casting the result of line 49 to an `int` would format the result nicely for the user ('85%' instead of '85.667%')

JLee35:
I like the idea of the new feature, but unfortunately it didn't work. One of your reviewers pointed this out, and it could have been easily fixed. It was also suggested that you write a test for your new functionality, and if you would have done so you probably would have caught that it wasn't working properly.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -11,20 +11,42 @@ public static void Main(string[] args)
             Question[] questions = LoadQuestions(filePath);
 
             int numberCorrect = 0;
+            //For loop for getting whether or not the user answers correctly
```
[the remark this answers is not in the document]
JLee35: Instead of a comment next time, try moving the code into it's own block with a descriptive name like `isUserRetrying()`

PR #41, opened 2023-01-23
Assignment2

jstanley6:


slanante:
Will do a review.

Assignment2/Logger.Tests/FileLoggerTests.cs
```diff
@@ -0,0 +1,28 @@
...
+    [TestClass]
+    public class FileLoggerTests
+    {
+        [TestMethod]
+        public void CreatesFileAndLogsMessage_Test()
+        {
+            string path = "test.txt";
+            FileLogger fileLogger = new FileLogger()
+            {
+                LogFilePath = path,
```
slanante, line 16: Compile Error here. Expects a semicolon.

slanante (commented):
Main concern for this pull review is in the organization of the repo. There are unnecessary files such as the PrincessBrideTrivia files and the .idea files and duplicates of directories for the Logger and Logger.Tests.

I added some code comments as well but the pull review's file organization added some difficulty in the review.

A cleaned up git push to this pull request would help ease future collaboration.
Attached is a screenshot of mentioned files with highlights on possible fixes:
image

Assignment2/Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+                streamWriter.Close();
+                
+            }
+            else
+            {
+                StreamWriter streamWriter = File.CreateText(LogFilePath);
+                streamWriter.Write(logMessage);
+                streamWriter.Close();
+            }
+        }
```
slanante, line 32: If the LogFilePath does not exist, do we want to create a new file by calling File.CreateText() in the else statement within the Log method? Perhaps we can call File.CreateText() within the FileLogger constructor.

Assignment2/Logger.Tests/FileLoggerTests.cs
jstanley6: There is a semi-colon at the end of the statement
   string path = "test.txt";
            FileLogger fileLogger = new FileLogger()
            {
                LogFilePath = path,
                ClassName = nameof(FileLoggerTests)

            }; <-----Not sure why it was giving a compile error for you, on mine it showed it successfully passed

slanante: Oops, the issue is on my end. An extension I have on VSCode was causing the issue.

Assignment2/.idea/.idea.Logger/.idea/.name
```diff
@@ -0,0 +1 @@
+Logger
```
JLee35, line 1: Good job adding a `.gitignore`, but the extensions specified in the ignore file will not stop tracking already tracked file extensions. After you add the ignore file, you would need to manually delete the files that you no longer want to track and add them back a second time for them to be ignored in source control.

Assignment2/Logger/LogFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+  
+    public class LogFactory
+    {
+        private string LogFilePath { get; set; }
+        public BaseLogger CreateLogger(string className)
+        {
+            if (!string.IsNullOrEmpty(LogFilePath))
+            {
+                BaseLogger baseLogger = new FileLogger(LogFilePath);
+                baseLogger.ClassName = className;
```
JLee35, line 13: While this still works, it doesn't meet the requirements of the writeup. The requirement was to use an **object initializer** which allows you to construct the object and assign `Classname` at the same time. Check out [these Microsoft docs for more details](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-initialize-objects-by-using-an-object-initializer)

```suggestion
                BaseLogger baseLogger = new FileLogger(LogFilePath) { Classname = classname }
```

Assignment2/Logger/LogFactory.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Logger
+{
+    
+  
+    public class LogFactory
+    {
+        private string LogFilePath { get; set; }
+        public BaseLogger CreateLogger(string className)
```
JLee35, line 8: Part of the assignment requirement was to enable `Nullable` and `TreatWarningsAsErrors` in `Logger.csproj`. If that would have been done, you would have seen that the code would have build errors forcing you to make the private string and the return type on the method nullable. Also remember that it is best practice to preface a private class field with an underscore (`_LogFilePath`)
```suggestion
        private string? _LogFilePath { get; set; }
        public BaseLogger? CreateLogger(string className)
```

Assignment2/Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            // Assert
+            Assert.AreEqual(1, logger.LoggedMessages.Count);
+            Assert.AreEqual(LogLevel.Error, logger.LoggedMessages[0].LogLevel);
+            Assert.AreEqual("Message 42", logger.LoggedMessages[0].Message);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentException))]
+        public void WarningWithNullLogger()
```
JLee35, line 39: Use descriptive names for your automated tests. As a best practice, they should follow the pattern of 'what is being tested', 'what is the test case', and 'what is the expected outcome':
```suggestion
        public void LoggerWarning_WithNullLogger_ThrowsException()
```

Assignment2/Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+        
+        [TestMethod]
+        public void DebugWithMessage()
+        {
+            var logger = new TestLogger();
+            logger.Debug("Message {0}", 42);
+            Assert.AreEqual(1, logger.LoggedMessages.Count);
+            Assert.AreEqual(LogLevel.Debug, logger.LoggedMessages[0].LogLevel);
+            Assert.AreEqual("Message 42", logger.LoggedMessages[0].Message);
```
JLee35, line 84: When writing tests, be sure to test for more than just the base expectation. We should have at least one test that checks that appending more than one log to a file actually contains all logs and doesn't overwrite previous logs.

Assignment2/Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Logger
+{
+    public static class BaseLoggerMixins
+    {
+        public static void Error(this BaseLogger baseLogger, string message, params object[] messageArgs)
+        {
+            if (baseLogger is null)
+            {
+                throw new ArgumentException("The BaseLogger can not be null.");
```
JLee35, line 11: Since `BaseLogger` is an abstract class, this really doesn't give the user the information it could. Instead, like the prompt says, we should use `nameof` here to get the name of the child logger class.
```suggestion
                throw new ArgumentException($"The {nameof(logger)} can not be null.");
```

Assignment2/Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+
+namespace Logger
+{
+    public static class BaseLoggerMixins
+    {
+        public static void Error(this BaseLogger baseLogger, string message, params object[] messageArgs)
```
JLee35, line 7: While the type of the first parameter is `BaseLogger`, the parameter will never actually be just a `BaseLogger` (because it is abstract and you can't instantiate an abstract class) but instead it will be a concrete class that inherits from `BaseLogger` like `FileLogger`. So `baseLogger` as a variable name is misleading.
```suggestion
        public static void Error(this BaseLogger logger, string message, params object[] messageArgs)
```

Assignment2/Logger/FileLogger.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public string LogFilePath { get; set; }
+
+        public FileLogger(){}
```
JLee35, line 10: Specifying a default constructor is unnecessary, it is implicitly given to you.
```suggestion
```

Assignment2/Logger/FileLogger.cs
```diff
@@ -0,0 +1,33 @@
...
+                StreamWriter streamWriter = File.AppendText(LogFilePath);
+                streamWriter.Write(logMessage);
+                streamWriter.Close();
+                
+            }
+            else
+            {
+                throw new ArgumentException("File Path is null");
+            }
+        }
```
JLee35, line 30: This isn't the functionality we want here. If the file did exist, then an `ArgumentException` would be thrown, but we would want to append our log to the end of the file. Since `File.AppendText` creates a file if one doesn't exist, you would want something like this: 
```suggestion
            var logMessage = DateTime.Now.ToString() + " " + ClassName + " " + logLevel + ": " + message + "\n";

            if (!string.IsNullOrEmpty(LogFilePath))
            {
                File.AppendAllText(LogFilePath, logMessage);
            }
            else
            {
                throw new ArgumentNullException($"{nameof(LogFilePath)} can not be null or empty.");
            }
        }
```

JLee35 (commented):
Reviewing this PR was difficult with the amount of unnecessary files checked in. Also, be sure to carefully read the assignment prompt.

PR #42, opened 2023-01-23
Assignment 2: Jacob Newcomb and Jessica Ho (No Extra-Credit)

jacobnewcomb:
with @JessicaH46

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -10,10 +10,9 @@ namespace Logger.Tests
         public void FileLogger_WritesToFile()
         {
             //Arrange
-            string className = "I should work!";
-            string fileName = "test.txt";
-            File.Delete(fileName);
-            FileLogger fileLogger = new FileLogger(className, fileName);
+            string className = nameof(FileLoggerTests);
+            string fileName = Path.GetTempFileName();
+            FileLogger fileLogger = new FileLogger(fileName) { ClassName = className};
 
             // Act
             fileLogger.Log(LogLevel.Debug, "I am debugingggggg mom");
@@ -23,10 +22,10 @@ namespace Logger.Tests
             string[] content = File.ReadAllLines(fileName);
             Assert.IsTrue(content.Length == 2);
             Assert.IsTrue(content[0].Contains("I am debugingggggg mom"));
-            Assert.IsTrue(content[0].Contains("I should work!"));
+            Assert.IsTrue(content[0].Contains(className));
             Assert.IsTrue(content[0].Contains("Debug: "));
             Assert.IsTrue(content[1].Contains("It's not a phase mom!"));
-            Assert.IsTrue(content[1].Contains("I should work!"));
+            Assert.IsTrue(content[1].Contains(className));
             Assert.IsTrue(content[1].Contains("Warning: "));
         }
     }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,33 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_WritesToFile()
+        {
+            //Arrange
+            string className = "I should work!";
+            string fileName = "test.txt";
+            File.Delete(fileName);
+            FileLogger fileLogger = new FileLogger(className, fileName);
+
+            // Act
+            fileLogger.Log(LogLevel.Debug, "I am debugingggggg mom");
+            fileLogger.Log(LogLevel.Warning, "It's not a phase mom!");
+
+            //Assert
+            string[] content = File.ReadAllLines(fileName);
+            Assert.IsTrue(content.Length == 2);
+            Assert.IsTrue(content[0].Contains("I am debugingggggg mom"));
+            Assert.IsTrue(content[0].Contains("I should work!"));
+            Assert.IsTrue(content[0].Contains("Debug: "));
+            Assert.IsTrue(content[1].Contains("It's not a phase mom!"));
+            Assert.IsTrue(content[1].Contains("I should work!"));
+            Assert.IsTrue(content[1].Contains("Warning: "));
+        }
     }
 }
--- a/Logger/BaseLogger.cs
+++ b/Logger/BaseLogger.cs
@@ -1,7 +1,11 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public abstract class BaseLogger
     {
+        private string? _className;
+        public string ClassName { get => _className; set => _className = value; }
         public abstract void Log(LogLevel logLevel, string message);
     }
 }
--- a/Logger/BaseLoggerMixins.cs
+++ b/Logger/BaseLoggerMixins.cs
@@ -5,80 +5,60 @@ namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        public static void Error(this BaseLogger baseLogger, string message, params object[] objects)
+        public static void Error(this BaseLogger baseLogger, string message, params object[] args)
         {
             if(baseLogger == null)
             {
                 throw new System.ArgumentNullException(nameof(baseLogger));
             }
-            if(objects != null)
+            if(args != null)
             {
-                StringBuilder sb = new StringBuilder(message);
-                for (int i = 0; i < objects.Length; i++)
-                {
-                    sb.Replace("{" + i + "}", objects[i].ToString());
-                }
-                baseLogger.Log(LogLevel.Error, sb.ToString());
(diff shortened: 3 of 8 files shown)


Wvvern:
Will Review

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.IO;
+using static System.Net.WebRequestMethods;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string className;
```
Wvvern, line 9: ClassName should be stored as a property within baselogger, and set by LogFactory during the instantiation of FileLogger using an object initializer.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+        public string fileName { get; set; }
+        public FileLogger(string className, string fileName) {
+            this.className = className;
+            this.fileName = fileName;
+        }
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = System.IO.File.AppendText(@fileName))
+            {
+                sw.WriteLine(DateTime.Now + " " + ClassName + " " + logLevel +": " + message);
```
Wvvern, line 20: You can use an interpolated string here instead of a bunch of concatenations.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.IO;
+using static System.Net.WebRequestMethods;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string className;
+        public string ClassName { get => className; set => className = value;}
```
Wvvern, line 10: Not sure why there is both a private field and property for ClassName

Logger/LogFactory.cs
```diff
@@ -2,10 +2,23 @@
 {
     public class LogFactory
     {
-        public BaseLogger CreateLogger(string className)
+        private string? fileName;
+        public string FileName { get => fileName; set => fileName = value; }
```
Wvvern, line 6: A property should be used here. The file path should be stored in a private field and set by ConfigureFileLogger.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,89 @@
...
+                StringBuilder sb = new StringBuilder(message);
+                for (int i = 0; i < objects.Length; i++)
+                {
+                    sb.Replace("{" + i + "}", objects[i].ToString());
+                }
+                baseLogger.Log(LogLevel.Error, sb.ToString());
+            }
+            else
+            {
+                baseLogger.Log(LogLevel.Error, message);
```
Wvvern, line 25:
```suggestion
                baseLogger.Log(LogLevel.Error, string.Format(message, args));
```
Not quite sure if that actually works, but a call to string.Format() should do the trick. 
Note, args is the existing parameter "objects". Just using a more clear name.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,33 @@
...
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_WritesToFile()
+        {
+            //Arrange
+            string className = "I should work!";
+            string fileName = "test.txt";
```
Wvvern, line 14: You could use a temp file here instead. Call Path.GetTempFileName() to get the file name.

Wvvern (commented):
## Assignment

- There is an existing `BaseLogger` class. It needs an **auto property** to hold a class name. This property should be set in the `LogFactory` using an **object initializer**. ❌
- Create a `FileLogger` that derives from `BaseLogger`. It should take in a path to a file to write the log message to. When its `Log` method is called, it should **append** messages on their own line in the file. The output should include all of the following:
  - The current date/time ✔
  - The name of the class that created the logger ✔
  - The log level ✔
  - The message ✔
  - The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
- The `LogFactory` should be updated with a new method `ConfigureFileLogger`. This should take in a file path and store it in a **private member**. It should use this when instantiating a new `FileLogger` in its `CreateLogger` method. ✔
- If the file logger has not be configured in the `LogFactory`, its `CreateLogger` method should return `null`. ✔
- Inside of `BaseLoggerMixins` implement **extension methods** on `BaseLogger` for
  - `Error`, ✔
  - `Warning`, ✔
  - `Information`, and ✔
  - `Debug`. ✔
  Each of these methods should take in a `string` for the message, as well as a **parameter array** of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the `BaseLogger.Log` method, by automatically supplying the appropriate `LogLevel`. These methods should throw an exception if the `BaseLogger` parameter is null. There are a couple example unit tests to get you started.
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional) ❌
- Turn on Nullability (`Nullable`) ✔
- **All of the above should be unit tested.**

## Extra Credit

- Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses `System.Console` or `System.Diagnostics.Trace` ❌

Logger/LogFactory.cs
```diff
@@ -2,10 +2,23 @@
...
         {
-            
-            return null;
+            if(string.IsNullOrEmpty(FileName))
+            {
+                return null;
+            }
+            else
+            {
+                FileLogger fl = new FileLogger(FileName, className);
```
Wvvern, line 15:
```suggestion
                FileLogger fl = new FileLogger(_filePath) { ClassName = className };
```
LogFactory should not be using a property for the file path (it should be a private field. An object initializer should also be used to set ClassName.

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,11 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public abstract class BaseLogger
     {
+        private string? _className;
+        public string ClassName { get => _className; set => _className = value; }
```
JLee35, line 8: While this works, it is not an **auto property** as is required in the assignment prompt. Using an **auto property** would look like this:
```suggestion
        public string ClassName { get; set; }
```
When using **auto-implemented properties** "the compiler creates a private, anonymous backing field that can only be accessed through the property's get and set accessors" [(see docs for more details).](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties) This means that declaring a private backing field and referencing that in your getters and setters is redundant.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.IO;
+using static System.Net.WebRequestMethods;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string _fileName;
+        public string FileName { get => _fileName; set => _fileName = value ?? throw new ArgumentNullException(value); }
```
JLee35, line 10: Adding a public setter for `_fileName` goes against what we want here. We are declaring `_fileName` private with the intention of protecting it from the rest of the world and only allowing it to be set within the current class. This is another perfect opportunity for an **auto implemented property** except this one should be readonly.
```suggestion
        public string FileName { get; }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public FileLogger(string fileName) {
+            FileName = fileName;
+        }
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = System.IO.File.AppendText(@FileName))
```
JLee35, line 17: What is the purpose of the '@' in front of `FileName`? It works fine without. `FileName` is not a reserved word, therefore this isn't necessary.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,69 @@
...
+        public static void Warning(this BaseLogger baseLogger, string message, params object[] args)
+        {
+            if (baseLogger == null)
+            {
+                throw new System.ArgumentNullException(nameof(baseLogger));
+            }
+            if (args != null)
+            {
+                baseLogger.Log(LogLevel.Warning, string.Format(new System.Globalization.CultureInfo("en-US"), message, args));
+            }
```
JLee35, line 32: Great check! Just be sure you have test coverage for it next time.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,32 @@
...
+            fileLogger.Log(LogLevel.Debug, "I am debugingggggg mom");
+            fileLogger.Log(LogLevel.Warning, "It's not a phase mom!");
+
+            //Assert
+            string[] content = File.ReadAllLines(fileName);
+            Assert.IsTrue(content.Length == 2);
+            Assert.IsTrue(content[0].Contains("I am debugingggggg mom"));
+            Assert.IsTrue(content[0].Contains(className));
+            Assert.IsTrue(content[0].Contains("Debug: "));
+            Assert.IsTrue(content[1].Contains("It's not a phase mom!"));
```
JLee35, line 27: 😆

JLee35 (commented):
Good job overall, just read up on **auto implemented properties** and be sure to check your `.csproj` file settings vs what is required in the assignment prompt.

PR #44, opened 2023-01-24
Assignment2: jdriscoll1 and alowe4

jdriscoll1:


Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,111 @@
...
+            
+            string message1 = "This is a test warning, please remain calm";
+            
+            string message2 = "This is a second test warning, please remain calm";
+
+            // Act
+            fileLogger.Log(LogLevel.Warning, message1);
+            fileLogger.Log(LogLevel.Warning, message2);
+
+            string expected1 = $"{DateTime.Now} FileLoggerTests Warning: {message1}";
```
jacobnewcomb, line 71: The only problem I can see with this is that DateTime.Now here and DateTime.Now in the Log method of the FileLogger could potentially be different because of latency.

jacobnewcomb (commented):
- There is an existing BaseLogger class. It needs an auto property to hold a class name. This property should be set in the LogFactory using an object initializer. ✔
- Create a FileLogger that derives from BaseLogger. It should take in a path to a file to write the log message to. When its Log method is called, it should append messages on their own line in the file. The output should include all of the following:
    The current date/time ✔
    The name of the class that created the logger ✔
    The log level ✔
    The message ✔
- The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ✔
- If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
    Inside of BaseLoggerMixins implement extension methods on BaseLogger for
    Error, ✔
    Warning, ✔
    Information, and ✔
    Debug. ✔
- Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔
- Turn on Nullability (Nullable) ✔

I think you hit all of the bases excluding extra credit! Everything looks great and it seems like you either know the language well or you did some good research. Either way, big ups!

lsobtafo (commented):
There is an existing BaseLogger class. It needs an auto property to hold a class name. This property should be set in the LogFactory using an object initializer. ✔
Create a FileLogger that derives from BaseLogger. It should take in a path to a file to write the log message to. When its Log method is called, it should append messages on their own line in the file. The output should include all of the following:
The current date/time ✔
The name of the class that created the logger ✔
The log level ✔
The message ✔
The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ✔
If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
Inside of BaseLoggerMixins implement extension methods on BaseLogger for
Error, ✔
Warning, ✔
Information, and ✔
Debug. ✔ 
Each of these methods should take in a string for the message, as well as a parameter array of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the BaseLogger.Log method, by automatically supplying the appropriate LogLevel. These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started.
Use the nameof() operator when identifying the class name to the logger ✔
Ensure you turn on Warnings as Errors (TreatWarningsAsErrors) ✔
Ensure that you turn on code analysis (EnableNETAnalyzers) ✔
Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
Turn on Nullability (Nullable) ✔
All of the above should be unit tested.

Extra Credit
Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses System.Console or System.Diagnostics.Trace ✔

No comments, great job!

Logger.Tests/FileLoggerTests.cs
jdriscoll1: Hey Jacob. Thanks for the comment. I agree that comparing two calls to DateTime.Now may cause issues due to latency. However, I'm not concerned about it because the parser runs through the program quickly enough that in only exceptional circumstances will this be an issue. However, I did consider this and I appreciate you leaving the feedback.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+            _CreationClass = creationClass;
+            _Filepath = filepath;
+            SetClassName(nameof(FileLogger));
+           
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            
+            string fullMessage = $" {DateTime.Now} {_CreationClass} {logLevel}: {message}".Trim();
```
mmwoodfo, line 26: What's the intended purpose of `Trim()` here? If the intention was to remove any white space at the start or end of `message`, I'd argue you want the trim to happen on the message variable it's self, not on the interpolated string.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+            
+            using (StreamWriter writer = File.AppendText(_Filepath))
+            {
+                writer.WriteLine(fullMessage);
+                writer.Flush();
+                writer.Close();
```
mmwoodfo, line 32: When using the `using` keyword, you should not need to close your writer. Once you exit scope, it will be auto cleaned up.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,36 @@
 {
     public class LogFactory
     {
-        public BaseLogger CreateLogger(string className)
+        private string _FilePath { get; set; } = null!;
```
mmwoodfo, line 5: Don't use `null!`, make the _FilePath nullable. Also, this is the syntax for an autoproperty:
`public string? FilePath {get; set;}`
This is the syntax for a private member
`private string? _FilePath;`

You're mixing the two styles together here.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,36 @@
...
+
+        public string GetFileLoggerPath() {
+            return _FilePath; 
+        }
+
+        public BaseLogger CreateLogger(string creationClass, string className)
         {
+
+
+            if (className == "ConsoleLogger")
```
mmwoodfo, line 20:
```suggestion
            if (className == "ConsoleLogger")
```
You should avoid using hard coded strings like this, use `nameof` instaed.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,36 @@
...
+            {
+                return new ConsoleLogger(creationClass);
+            }
+            if (_FilePath is null) {
```
mmwoodfo, line 24: This should be your first check.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,36 @@
...
+            return _FilePath; 
+        }
+
+        public BaseLogger CreateLogger(string creationClass, string className)
         {
+
+
+            if (className == "ConsoleLogger")
+            {
+                return new ConsoleLogger(creationClass);
```
mmwoodfo, line 22: This should probably take a file path, rather than the creating class's name.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,36 @@
...
+                return null!; 
+            }
+
+            if (className == "FileLogger")
+            {
+                return new FileLogger(creationClass, _FilePath);
+            }
             
-            return null;
+            return null!;
```
mmwoodfo, line 33: Do not use `null!`, make your method signature nullable.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,36 @@
...
+
+        public void ConfigureFileLogger(string fileLogger) {
+            _FilePath = fileLogger; 
+        }
+
+        public string GetFileLoggerPath() {
+            return _FilePath; 
+        }
+
+        public BaseLogger CreateLogger(string creationClass, string className)
```
mmwoodfo, line 16:
```suggestion
        public BaseLogger? CreateLogger(string creationClass, string className)
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,67 @@
...
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+
+       
+        public static void Error(this BaseLogger logger, string msg, params string[] args)
+        {
+            if (msg is null) {
```
mmwoodfo, line 11: You aren't reuquired to do this check.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,67 @@
...
-        
+
+       
+        public static void Error(this BaseLogger logger, string msg, params string[] args)
+        {
+            if (msg is null) { 
+                throw new ArgumentNullException(msg);
+            }
+            if (logger is null) {
+                throw new ArgumentNullException(logger!.ToString());
```
mmwoodfo, line 15:
```suggestion
                throw new ArgumentNullException(nameof(logger));
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,67 @@
...
+            }
+            String errorMessage = string.Format(null, msg, args);
```
mmwoodfo, line 17:
```suggestion
            String errorMessage = string.Format(msg, args);
```

PR #47, opened 2023-01-24
Assignment 2

TriciaFaust:


Logger.Tests/LogFactoryTests.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        [TestMethod]
+        public void LogFactoryTest()
+        {
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger("C:\\Logs");
+            var fileLogger = logFactory.CreateLogger(nameof(LogFactoryTests));
+            Assert.IsNotNull(fileLogger);
+            Assert.AreEqual(nameof(LogFactoryTests), fileLogger.Name);
+            Assert.AreEqual("C:\\Logs\\LogFactoryTests.log", fileLogger.PathName);
```
Bhopss152, line 17: Hard coding a path is not the best since if another user want to use it on their pc it wouldnt work right.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    [TestClass]
+    public class FileLoggerTests
+    {
+        BaseLogger fileLogger;
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger("C:\\Logs");
```
Bhopss152, line 13: Same thing here with the path.

Logger.Tests/LogFactoryTests.cs
GeorgeKoralski007, line 17: To add to what Bhopss152 said, I would make a Tempfile and then have the file in the Tempfile deleted so it does not overwrite itself.

Logger/BaseLogger.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Logger
+{
+    public abstract class BaseLogger
+    {
+        public abstract void Log(LogLevel logLevel, string message);
+
+        public string Name { get; set; }
+        public string PathName { get; set; }
```
GeorgeKoralski007, line 8: PathName should be defined in FileLogger class. Base class does not need it because its not common for all loggers.

Logger/LogFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+        public BaseLogger CreateLogger(string className)
+        {
+            ClassName = className;
+            if (OutputPath == null)
+            {
+                return null;
+            }
+
+            string outputPath = Path.Combine(OutputPath, ClassName + ".log");
+            FileLogger fileLogger = new FileLogger(outputPath);
```
GeorgeKoralski007, line 20: FileLogger should only be created when className parameter is "FileLogger"
Otherwise it should return null

Logger/LogFactory.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.IO;
+using System.Xml.Linq;
+using System.Xml.XPath;
+
+namespace Logger
+{
+    public class LogFactory
+    {
+        private string ClassName;
```
GeorgeKoralski007, line 9: This is not needed

Logger/LogFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            ClassName = className;
+            if (OutputPath == null)
+            {
+                return null;
+            }
+
+            string outputPath = Path.Combine(OutputPath, ClassName + ".log");
+            FileLogger fileLogger = new FileLogger(outputPath);
+            fileLogger.Name = ClassName;
```
GeorgeKoralski007, line 21: Remove the ClassName and use only the parameter className here

Logger/LogFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+        private string OutputPath;
+        public BaseLogger CreateLogger(string className)
+        {
+            ClassName = className;
+            if (OutputPath == null)
+            {
+                return null;
+            }
+
+            string outputPath = Path.Combine(OutputPath, ClassName + ".log");
```
GeorgeKoralski007, line 19: Check if OutputPath is previously set. If still null then return null because we cannot create FileLogger.
OutputPath should have the filename in it also (not only folder)

Logger.Tests/LogFactoryTests.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    [TestClass]
+    public class LogFactoryTests
+    {
+        [TestMethod]
+        public void LogFactoryTest()
+        {
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger("C:\\Logs");
+            var fileLogger = logFactory.CreateLogger(nameof(LogFactoryTests));
```
GeorgeKoralski007, line 14: It should use nameof(FileLogger)

Logger/LogFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+            FileLogger fileLogger = new FileLogger(outputPath);
+            fileLogger.Name = ClassName;
+            return fileLogger;
+        }
+
+        public void ConfigureFileLogger(string pathName) 
+        {
+            OutputPath = pathName;
```
GeorgeKoralski007, line 27: This should be the full file path including the file name.
Currently you are using it as a directory path only.

GeorgeKoralski007 (changes requested):

mmwoodfo (commented):
image


Make sure your PR is going against the correct branch, in this case, it should be going against the Assignment 2 branch.

PR #48, opened 2023-01-24
Assignment 2, joshmerrell and Nolan-Posey

joshmerrell:
(Joshua Merrell and Nolan Posey)
Extra credit attempted, using System.Console


BenFristadSchool (commented):
I will review this

Logger/LogFactory.cs
```diff
@@ -1,10 +1,30 @@
-﻿namespace Logger;
+﻿using System;
+
+namespace Logger;
 
 public class LogFactory
 {
-    public BaseLogger CreateLogger(string className)
+    private string LogFilePath = "";
```
BenFristadSchool: You don't need to set the field to empty string. You can just leave it declared without setting it and then set it from within configureFileLogger using an auto property. Also proper convention is to set you field to the name of your property with Pascal case and a underscore `private string FilePath_`

Logger/LogFactory.cs
```diff
@@ -1,10 +1,30 @@
...
+    public void ConfigureLogger(string filePath)
```
BenFristadSchool: Should change to auto property

Logger/Logger.csproj
```diff
@@ -2,7 +2,16 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>10.0</LangVersion>
+    <LangVersion>11.0</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
BenFristadSchool, line 11: what is the reason for nested PropertyGroup tags?

Logger/LogFactory.cs
```diff
@@ -1,10 +1,30 @@
...
+    {
+        LogFilePath = filePath;
+    }
+    public BaseLogger? CreateLogger(string className)
     {
-        
-        return null;
+        if (LogFilePath == null)
```
BenFristadSchool: you can use `is` keyword rather than `==`

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+                break;
+            case LogLevel.Information:
+                msgOut += "Information ";
+                break;
+            case LogLevel.Debug:
+                msgOut += "Debug ";
+                break;
+        }
+
+        LogFile.WriteLine(msgOut + message);
```
BenFristadSchool: You could also use string interpolation here rather than concatenating strings and you can remove the switch statement 
Ex:  `LogFile.WriteLine($"{DateTime.Now} {ClassName} {LogLevel}: {message}");`

BenFristadSchool (commented):
Where is FileLoggerTests.cs? You want to ensure that your logger is producing proper output.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
-    }
-}
+﻿using System;
+using System.Diagnostics;
+
+namespace Logger;
+
+public class LogFactory
+{
+    private string LogFilePath = "";
```
BenFristadSchool, line 8: You don't need to set the field to empty string. You can just leave it declared without setting it and then set it from within configureFileLogger using an auto property. Also proper convention is to set you field to the name of your property with Pascal case and a underscore `private string FilePath_`

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
+    public void ConfigureLogger(string filePath)
```
BenFristadSchool, line 9: Should change to auto property

Logger/FileLogger.cs
```diff
@@ -0,0 +1,34 @@
...
+                break;
+            case LogLevel.Information:
+                msgOut += "Information";
+                break;
+            case LogLevel.Debug:
+                msgOut += "Debug";
+                break;
+        }
+
+        LogFile.WriteLine(msgOut + "] " + message);
```
BenFristadSchool, line 30: You could also use string interpolation here rather than concatenating strings and you can remove the switch statement 
Ex:  `LogFile.WriteLine($"{DateTime.Now} {ClassName} {LogLevel}: {message}");`

BenFristadSchool (commented):
Well Done!

Logger/Logger.csproj
joshmerrell: If you're referring to the duplicate PropertyGroup with the Condition attributes, I think that may have been added in by Visual Studio at some point by mistake, and I never removed it, just in case that might break something.

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,9 @@
 
 public abstract class BaseLogger
 {
+    public string? ClassName { get; set; }
+    public string? CallingClass { get; set; }
```
mmwoodfo, line 6: What is the difference between ClassName and CallingClass here?

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,44 @@
...
+namespace Logger;
 
 public static class BaseLoggerMixins
 {
-    
+    public static void Error(this BaseLogger log, string message, params object[] args)
+    {
+        if (log is null)
+        {
+            throw new ArgumentNullException("log must not be null");
```
mmwoodfo, line 12:
```suggestion
            throw new ArgumentNullException(nameof(log));
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,44 @@
-﻿namespace Logger;
+﻿using System;
+using System.Linq;
+
+namespace Logger;
 
 public static class BaseLoggerMixins
 {
-    
+    public static void Error(this BaseLogger log, string message, params object[] args)
```
mmwoodfo, line 8:
```suggestion
    public static void Error(this BaseLogger logger, string message, params object[] args)
```
A better name here would be 'logger' since that's what you're passing in, a type of logger

Logger/ConsoleLogger.cs
```diff
@@ -0,0 +1,27 @@
...
+                break;
+            case LogLevel.Information:
+                msgOut += "Information";
+                break;
+            case LogLevel.Debug:
+                msgOut += "Debug";
+                break;
+        }
+
+        Console.WriteLine(msgOut + "] " + message);
```
mmwoodfo, line 25: Consider using string interpolation here
```suggestion
        Console.WriteLine($"{msgOut}] {message}");

```

mmwoodfo: Additionally, you may consider getting the log level and setting that to a variable and then building your string message, to improve readability, rather than having the switch statement in the middle and appending to your string within your switch.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.IO;
+
+namespace Logger;
+internal class FileLogger : BaseLogger
+{
+    private readonly StreamWriter LogFile;
+    public FileLogger(string logPath) {
+        LogFile = new StreamWriter(logPath, true);
+    }
```
mmwoodfo, line 10:
```suggestion
    private readonly string _FilePath;
    public FileLogger(string filePath) {
        _FilePath= filePath;
    }
```

I would not recommend setting up a StreamWriter here. I'd recommend only using your stream writer in your log method and then disposing of it. Doing it this way can create concurrency issues.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
+using System.Diagnostics;
+
+namespace Logger;
+
+public class LogFactory
+{
+    public string? LogFilePath { get; set; }
+    public void ConfigureLogger(string filePath)
+    {
+        LogFilePath = filePath;
```
mmwoodfo, line 11: Try to keep your naming consistent for your variables.
```suggestion
        _FilePath = filePath;

```

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
-    }
-}
+﻿using System;
+using System.Diagnostics;
+
+namespace Logger;
+
+public class LogFactory
+{
+    public string? LogFilePath { get; set; }
```
mmwoodfo, line 8:
```suggestion
    private string? _FilePath;
```
This should be private and is not an auto property, just a field

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
+    public void ConfigureLogger(string filePath)
+    {
+        LogFilePath = filePath;
+    }
+    public BaseLogger? CreateLogger(string className, string callingClass)
+    {
+        BaseLogger? log;
+        switch (className)
+        {
+            case "ConsoleLogger":
```
mmwoodfo, line 18: Use nameof here
```suggestion
            case nameof(ConsoleLogger):

```

Logger/LogFactory.cs
```diff
@@ -1,10 +1,34 @@
...
+                log = new ConsoleLogger();
+                log.ClassName = className;
+                break;
+            case "FileLogger":
+                if (LogFilePath is null)
+                    return null;
+                log = new FileLogger(LogFilePath);
+                log.ClassName = className;
```
mmwoodfo, line 26:
```suggestion
                log = new FileLogger(LogFilePath)
                {
                    ClassName = className;
                }
```
This is how you use an object initializer

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,51 @@
...
+[TestClass]
+public class FileLoggerTests
+{
+    [TestMethod]
+    public void FileLogger_IsNull_ReturnsTrue()
+    {
+        //Arrange
+        LogFactory factory = new();
+        factory.ConfigureLogger(null);
+        BaseLogger fileLogger = factory.CreateLogger("FileLogger", nameof(FileLoggerTests));
```
mmwoodfo, line 14: This should be under your `//Act` section

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,51 @@
...
-}
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class FileLoggerTests
+{
+    [TestMethod]
+    public void FileLogger_IsNull_ReturnsTrue()
```
mmwoodfo, line 9:
```suggestion
    public void FileLogger_FileLoggerNotConfiguredWithPath_CreateLoggerReturnsTrue()

```
Consider a more explicit name for your tests, something like this maybe?

PR #50, opened 2023-01-24
Assignment 2 - Hunter Thompson & Hunter Allen

StrangeRanger:
@hallen10 (Hunter Allen)

---

~~***IMPORTANT***: THE PATH TO THE OUTPUT TEXT FILE USES THE UNIX FILE SYSTEM FORMAT, INSTEAD OF THE WINDOWS `C:\BLAH\BLAH` FORMATTING. YOU'LL EITHER WANT TO CHANGE IT TO MEET YOUR OPERATING SYSTEM NEEDS OR EXECUTE IT ON LINUX OR MACOS.~~


Wvvern:
Will review

Wvvern (commented):
## Assignment

- There is an existing `BaseLogger` class. It needs an **auto property** to hold a class name. This property should be set in the `LogFactory` using an **object initializer**. ❌
- Create a `FileLogger` that derives from `BaseLogger`. It should take in a path to a file to write the log message to. When its `Log` method is called, it should **append** messages on their own line in the file. The output should include all of the following:
  - The current date/time ✔
  - The name of the class that created the logger ❌
  - The log level ✔
  - The message ✔
  - The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
- The `LogFactory` should be updated with a new method `ConfigureFileLogger`. This should take in a file path and store it in a **private member**. It should use this when instantiating a new `FileLogger` in its `CreateLogger` method. ✔
- If the file logger has not be configured in the `LogFactory`, its `CreateLogger` method should return `null`. ✔
- Inside of `BaseLoggerMixins` implement **extension methods** on `BaseLogger` for
  - `Error`, ✔?
  - `Warning`, ✔?
  - `Information`, and ✔?
  - `Debug`. ✔?
  Each of these methods should take in a `string` for the message, as well as a **parameter array** of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the `BaseLogger.Log` method, by automatically supplying the appropriate `LogLevel`. These methods should throw an exception if the `BaseLogger` parameter is null. There are a couple example unit tests to get you started.
- Turn on Nullability (`Nullable`) ✔
- **All of the above should be unit tested.**

## Extra Credit

- Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses `System.Console` or `System.Diagnostics.Trace` ❌

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,38 @@
-﻿namespace Logger;
+using System;
+namespace Logger;
 
 public static class BaseLoggerMixins
 {
-    
+    public static void Error(this BaseLogger baseLogger, string message, params int[] args)
```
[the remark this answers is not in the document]
StrangeRanger: Why should it be typeless? We are explicitly dealing with integers, so would it not be preferable to specify int?

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,38 @@
...
+    {
+        if (baseLogger is null) { throw new ArgumentNullException(); }
+
+        LogLevel logLevel = LogLevel.Error;
+        baseLogger.Log(logLevel, string.Format(message, args[0]));
```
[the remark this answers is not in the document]
StrangeRanger: using args results in the producing "Actual: <Message System.Int32[]>", instead of the expected Message 42

Logger/BaseLoggerMixins.cs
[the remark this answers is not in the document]
Wvvern: https://i.imgur.com/NedINMc.png
Although the tests only pass in an int, there's no reason to limit the interpolation to ints only. 
If you take a look at the passing test in the image above, my understanding of the assignment is that there could be several parameters passed in for interpolation, and each parameter could also have its own unique type.

Logger/BaseLoggerMixins.cs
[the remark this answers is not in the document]
Wvvern: I think that's because there is no String.Format(String, Int[]) method in c#, but there is a string.Format(String, Object).
Even though String.Format(String, Object[]) exists, for some reason, when not given a generic object array, it just interpolates the object type as the string representation.
To resolve this issue, set args to an array of objects (typeless) - see other comment. This also allows us to pass in more than one parameter to the interpolated string - think of printf() in c.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,27 @@
...
+    
+    public FileLogger(string filePath)
+    {
+        FilePath = filePath;
+    }
+
+    public override void Log(LogLevel logLevel, string message)
+    {
+        DateTime date = DateTime.Now;
+        ClassName     = nameof(FileLogger);
```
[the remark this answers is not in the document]
StrangeRanger: If this is left out, while no error will be thrown, nothing for the ClassName will be printed into the file. So this is necessary.

Logger/BaseLoggerMixins.cs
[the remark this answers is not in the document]
StrangeRanger: Ah, I see. Yea, it seemed to work once I used object instead of int. Thanks

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,23 @@
...
+using System;
+using System.IO;
 namespace Logger.Tests;
 
 [TestClass]
 public class FileLoggerTests
 {
-    
+    [TestMethod]
+    public void LogTest()
```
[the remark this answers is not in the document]
StrangeRanger: From my understanding, the classname doesn't need to be called or specified here. We expect it to be FilePath, so we hardcode it into the assert. Therefore it isn't necessary to set ClassName.

Logger/FileLogger.cs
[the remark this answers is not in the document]
Wvvern: I think the idea is that ClassName can be set to anything, and not just to FileLogger. So whichever class is attempting to do logging should pass its own name into ClassName. Now that LogFactory.CreateLogger() properly uses it's className parameter to set the ClassName property, I don't think that should be happening. Have you checked after updating? Within FileLoggerTests.cs, set the ClassName of the file logger, and check for the existance of the ClassName within the logged text.

Logger.Tests/FileLoggerTests.cs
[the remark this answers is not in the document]
Wvvern: See other comment, the inherited ClassName property of FileLogger should be set and checked for within the logged text.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.IO;
+namespace Logger;
+
+public class FileLogger : BaseLogger
+{
+    // Question: What is the purpose are the auto property? Should it always be public or is
+    //           it ok to make it private.
```
mmwoodfo, line 8: In general, auto properties should be public and fields should be private and not the other way around. This is mostly just a coding convention.

The .NET framework assumes that you will use properties instead of public fields, properties allow you to implement validation, use separate accessibility levels for the getter and setter (example: `public string FilePath { get; private set; }`), and it's difficult to change your public fields into properties in the future, so it's better to start out with them.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+    private string FilePath { get; }
```
mmwoodfo, line 9:
```suggestion
    private string _FilePath;
```

Logger/LogFactory.cs
```diff
@@ -1,10 +1,17 @@
-﻿namespace Logger;
+namespace Logger;
 
 public class LogFactory
 {
+    private string? _filePath;
```
mmwoodfo, line 5:
```suggestion
    private string? _FilePath;
```
For this class, you should be doing pascal casing for your fields. You will find that this guideline is highly debated in the community.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,17 @@
...
+
     public BaseLogger CreateLogger(string className)
     {
-        
-        return null;
+        if (_filePath is null) { return null!; }
+        return new FileLogger(_filePath, className);
```
mmwoodfo, line 10: Instead of passing through the constructor, you should be using an object initializer 
```suggestion
        return new FileLogger(_filePath,)
        {
            ClassName = className
        }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public FileLogger(string filePath, string className)
```
mmwoodfo, line 11:
```suggestion
    public FileLogger(string filePath)
```
Per the assignment instructions, this shouldn't be passed in as a constructor parameter, you should set this via object initialization (I left a comment on this in your logFactory)

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        ClassName = className;
+        FilePath  = filePath;
+    }
+
+    public override void Log(LogLevel logLevel, string message)
+    {
+        DateTime date     = DateTime.Now;
+        string   outGoing = $"{date} {ClassName} {logLevel}: {message}";
```
mmwoodfo, line 20: You have some weird spacing going on here

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        StreamWriter streamWriter = File.CreateText(FilePath);
```
mmwoodfo, line 22: Consider using a `using` statement here.

Logger/LogFactory.cs
```diff
@@ -1,10 +1,17 @@
...
 
 public class LogFactory
 {
+    private string? _filePath;
+
     public BaseLogger CreateLogger(string className)
     {
-        
-        return null;
+        if (_filePath is null) { return null!; }
```
mmwoodfo, line 9:
```suggestion
        if (_filePath is null) { return null; }
```
avoid using `!` whenever possible, if you find yourself using this, ask yourself if there is a better solution. In this case, the better solution is to change your method signature to return a nullable `BaseLogger` on line 5. 

Like this:
`public BaseLogger? CreateLogger(string className)`

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,39 @@
-﻿namespace Logger;
+using System;
+namespace Logger;
 
 public static class BaseLoggerMixins
 {
-    
+    public static void Error(this BaseLogger baseLogger, string message, params object[] args)
+    {
+        if (baseLogger is null) { throw new ArgumentNullException(); }
```
mmwoodfo, line 8:
```suggestion
        if (baseLogger is null) 
        { 
            throw new ArgumentNullException(); 
        }
```
OR 
```suggestion
        if (baseLogger is null) throw new ArgumentNullException();
```
Although the second suggestion has some strong opinions around not doing that, the first suggestion is the safer bet

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,39 @@
...
+
+        LogLevel logLevel = LogLevel.Warning;
+        baseLogger.Log(logLevel, string.Format(message, args));
+    }
+
+    public static void Information(this BaseLogger baseLogger, string message,
+                                   params object[] args)
+    {
+        if (baseLogger is null) { throw new ArgumentNullException(); }
```
mmwoodfo, line 26:
```suggestion
        if (baseLogger is null) { throw new ArgumentNullException(nameof(baselogger)); }
```
Make sure in all of your throws you are telling the exception what argument is unexpectedly null

Logger/BaseLoggerMixins.cs
```diff
@@ -1,6 +1,39 @@
...
+    public static void Warning(this BaseLogger baseLogger, string message,
+                               params object[] args)
+    {
+        if (baseLogger is null) { throw new ArgumentNullException(); }
+
+        LogLevel logLevel = LogLevel.Warning;
+        baseLogger.Log(logLevel, string.Format(message, args));
+    }
+
+    public static void Information(this BaseLogger baseLogger, string message,
```
mmwoodfo, line 23:
```suggestion
    public static void Information(this BaseLogger logger, string message,
```
I'd recommend calling this 'logger' as your BaseLogger is a base class with lots of children that can be inheriting from it, so, this could actually be a FileLogger for example.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,9 +1,25 @@
...
 namespace Logger.Tests;
 
 [TestClass]
 public class FileLoggerTests
 {
-    
+    [TestMethod]
+    public void LogTest_TestMessage_AreEqual()
+    {
+        string       path         = Path.GetTempFileName();
```
mmwoodfo, line 11: Don't add this extra space padding around your variables like this, you're making it look like PowerShell. 




I don't like PowerShell.

StrangeRanger: I space them out so adjacent variables are aligned with each. This makes it more readable, in my opinion. Is there a good reason for why I shouldn't do this?

Logger/BaseLoggerMixins.cs
StrangeRanger: If I have the space, I will prefer to put it on one line. But I always use braces to remain syntactically consistent.

Logger/FileLogger.cs
StrangeRanger: I space them out so adjacent variables are aligned with each. This makes it more readable, in my opinion.

Logger.Tests/FileLoggerTests.cs
StrangeRanger: What if I instead of aligning both the variable and assignment operator, I only aligned the assignment operator? Would that be ok instead?

StrangeRanger (commented):
@mmwoodfo

Logger.Tests/FileLoggerTests.cs
mmwoodfo: Yeah, this is pretty atypical in C#. I think something similar was mentioned in class on Tuesday. There are certain conventions that are different in certain languages. I'd argue, PowerShell does something like this when declaring hash tables (not regular variables), but you would never see something like this in C#.

One of the biggest reasons we have guidelines and best practices is so that when there are multiple developers working on the same project, all the code looks the same. If you and I were working on a project together, we'd have to agree on how to format the code and what standards we'd want to set up to make sure our code looks the same. Usually, those standards are decided for the whole community of developers for a particular language, that way you aren't changing standards between code bases for the same language. 

Going against convention decreases readability for those who are familiar with the conventions for the language, and causes problems if we were to introduce a linter to your code base (a linter is something that auto formats your code and does basic error checking), and causes your code to be different than other C# code.

TL;DR this goes against convention and it will cause your code to be mismatched against how other developers in the C# community write their code.

PR #55, opened 2023-01-24
Assignment2 - Ian Kaiserman

Afterlyte:
Paired with @eroadhouse (Evan Rohde)

I made one commit this morning to fix the issues with the changes being displayed since I had manually changed the files to file-scoped namespaces, they are now back to block-scoped like the original so that GitHub displays the changes more clearly.

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,25 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void Log_GivenLog_Success()
+        {
+            LogFactory testFactory = new();
+            testFactory.ConfigureFileLogger("test.txt");
+            FileLogger testFileLogger = (FileLogger)testFactory.CreateLogger(nameof(FileLoggerTests));
+            testFileLogger.Information("Test message {0}", "42");
+
+            string[] fileText = File.ReadAllLines(testFileLogger.FilePath);
+            string newestLine = fileText[^1];
+
+            Assert.AreEqual(string.Format($"{DateTime.Now} {testFileLogger.ClassName} {LogLevel.Information}: Test message 42"), newestLine);
+
+        }
     }
 }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -2,24 +2,23 @@
 using System;
 using System.IO;
 
-namespace Logger.Tests
+namespace Logger.Tests;
+
+[TestClass]
+public class FileLoggerTests
 {
-    [TestClass]
-    public class FileLoggerTests
+    [TestMethod]
+    public void Log_GivenLog_Success()
     {
-        [TestMethod]
-        public void Log_GivenLog_Success()
-        {
-            LogFactory testFactory = new();
-            testFactory.ConfigureFileLogger("test.txt");
-            FileLogger testFileLogger = (FileLogger)testFactory.CreateLogger(nameof(FileLoggerTests));
-            testFileLogger.Information("Test message {0}", "42");
+        LogFactory testFactory = new();
+        testFactory.ConfigureFileLogger("test.txt");
+        FileLogger testFileLogger = (FileLogger)testFactory.CreateLogger(nameof(FileLoggerTests));
+        testFileLogger.Information("Test message {0}", "42");
 
-            string[] fileText = File.ReadAllLines(testFileLogger.FilePath);
-            string newestLine = fileText[^1];
+        string[] fileText = File.ReadAllLines(testFileLogger.FilePath);
+        string newestLine = fileText[^1];
 
-            Assert.AreEqual(string.Format($"{DateTime.Now} {testFileLogger.ClassName} {LogLevel.Information}: Test message 42"), newestLine);
+        Assert.AreEqual(string.Format($"{DateTime.Now} {testFileLogger.ClassName} {LogLevel.Information}: Test message 42"), newestLine);
 
-        }
     }
 }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -10,14 +10,17 @@ public class FileLoggerTests
     [TestMethod]
     public void Log_GivenLog_Success()
     {
+        // Arrange
         LogFactory testFactory = new();
         testFactory.ConfigureFileLogger("test.txt");
         FileLogger testFileLogger = (FileLogger)testFactory.CreateLogger(nameof(FileLoggerTests));
-        testFileLogger.Information("Test message {0}", "42");
 
+        // Act
+        testFileLogger.Information("Test message {0}", "42");
         string[] fileText = File.ReadAllLines(testFileLogger.FilePath);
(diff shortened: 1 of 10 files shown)


Clockknight:
I will review this

Clockknight:
There is an existing BaseLogger class. It needs an auto property to hold a class name. This property should be set in the LogFactory using an object initializer. ✔
Create a FileLogger that derives from BaseLogger. It should take in a path to a file to write the log message to. When its Log method is called, it should append messages on their own line in the file. The output should include all of the following:
The current date/time ✔
The name of the class that created the logger ✔
The log level ✔
The message ✔
The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ✔
If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
Inside of BaseLoggerMixins implement extension methods on BaseLogger for
Error, ✔
Warning, ✔
Information, and ✔
Debug. ✔ Each of these methods should take in a string for the message, as well as a parameter array of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the BaseLogger.Log method, by automatically supplying the appropriate LogLevel. These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started.
Use the nameof() operator when identifying the class name to the logger ✔
Ensure you turn on Warnings as Errors (TreatWarningsAsErrors) ✔
Ensure that you turn on code analysis (EnableNETAnalyzers) ✔
Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
Turn on Nullability (Nullable) ✔
All of the above should be unit tested.

mmoran314 (commented):
There is an existing BaseLogger class. It needs an auto property to hold a class name. This property should be set in the LogFactory using an object initializer. ✔
Create a FileLogger that derives from BaseLogger. It should take in a path to a file to write the log message to. When its Log method is called, it should append messages on their own line in the file. The output should include all of the following:
The current date/time ✔
The name of the class that created the logger ✔
The log level ✔
The message ✔
The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"
The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ✔
If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
Inside of BaseLoggerMixins implement extension methods on BaseLogger for
Error, ✔
Warning, ✔
Information, and ✔
Debug. ✔ Each of these methods should take in a string for the message, as well as a parameter array of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the BaseLogger.Log method, by automatically supplying the appropriate LogLevel. These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started.
Use the nameof() operator when identifying the class name to the logger ✔
Ensure you turn on Warnings as Errors (TreatWarningsAsErrors) ✔
Ensure that you turn on code analysis (EnableNETAnalyzers) ✔
Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
Turn on Nullability (Nullable) ✔
All of the above should be unit tested.

Looks good to me!
The only thing I saw which may only be because I misinterpreted the assignment is writing the message of each log on a separate line from the rest of the info rather than putting everything on one line each log. That may only be from my interpretation of that part of the assignment.
Other than that it looks good to me!

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,28 @@
...
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void Log_GivenLog_Success()
```
JLee35, line 11: You've got the naming convention correct, but we should be more explicit with our expected result. `Success` is a very general idea, and could be up for interpretation. 
```suggestion
        public void Logger_Created_AppendsMessageToFile()
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,28 @@
...
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger logger, string message, params Object[] args)
+        {
+            if (logger is null) { throw new ArgumentNullException(nameof(logger)); }
+            else { logger.Log(LogLevel.Error, string.Format(message, args)); }
```
JLee35, line 10: `else` is unnecessary.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public FileLogger(string filePath)
+        {
+            FilePath = filePath;
+        }
+
+        public string FilePath { get; private set; }
```
JLee35, line 13: Prompt called for `FilePath` to be private.
```suggestion
        private string FilePath { get; set; }
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,30 @@
...
+        {
+            get
+            {
+                return _FileLoggerPath;
+            }
+            private set
+            {
+                _FileLoggerPath = value;
+            }
+        }
```
JLee35, line 16: No need for a public field with a private backing.
```suggestion
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,30 @@
...
+
+        public void ConfigureFileLogger(string filePath)
+        {
+            FileLoggerPath = filePath;
```
JLee35, line 20:
```suggestion
            _FileLoggerPath = filePath;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,30 @@
...
+        }
+
         public BaseLogger CreateLogger(string className)
         {
-            
-            return null;
+            if (FileLoggerPath is null) { return null!; }
```
JLee35, line 25:
```suggestion
            if (_FileLoggerPath is null) { return null; }
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,30 @@
...
+                _FileLoggerPath = value;
+            }
+        }
+
+        public void ConfigureFileLogger(string filePath)
+        {
+            FileLoggerPath = filePath;
+        }
+
         public BaseLogger CreateLogger(string className)
```
JLee35, line 23:
```suggestion
        public BaseLogger? CreateLogger(string className)
```

JLee35 (commented):
Good job.

Logger/LogFactory.cs
Afterlyte: > No need for a public field with a private backing.

Isn't it conventional to not modify member fields outside of wrapping properties? I'm confused as to why I would remove this and change the references to the underscored field name, or am I missing something?

JLee35: The idea of encapsulation is to hide class internals from outside classes. By adding a public field for getting the private field, you are exposing that field to the world. We don't want that here (but you might in some cases). Second, in this scenario we only want to be able to modify `_FileLoggerPath` from within the constructor, but the way you have it can be edited anywhere from within the class (even after instantiation). We want to be as restrictive as possible, thus in this case, we only want a private field that is set only in the constructor (so make it readonly).

PR #63, opened 2023-01-27
Assignment2

lsobtafo:


Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Logger;
+
+public static class BaseLoggerMixins
+{
+    public static void Error(this BaseLogger baseLogger, string message, params object[] args)
+   { 
+        if (baseLogger == null) { throw new ArgumentNullException(nameof(baseLogger)); }
```
mmwoodfo, line 9: This should be multiple lines
```suggestion
        if (baseLogger is null) 
        { 
           throw new ArgumentNullException(nameof(baseLogger)); 
        }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Logger;
+
+public static class BaseLoggerMixins
+{
+    public static void Error(this BaseLogger baseLogger, string message, params object[] args)
```
mmwoodfo, line 7:
```suggestion
    public static void Error(this BaseLogger logger, string message, params object[] args)
```
Rename your variable to `logger` instead of `baseLogger` because this variable won't always, directly, be a BaseLogger... It could be a FileLogger which is a type of BaseLogger

mmwoodfo:
image

This PR is going against the main branch, make sure you are going against the correct assignment branch.

PR #64, opened 2023-01-29
Assignment 3 @avanchu253

randypipkins:
Plus extra credit 2 completed

### DIFF
--- a/CanHazFunny/CanHazFunny/Jester.cs
+++ b/CanHazFunny/CanHazFunny/Jester.cs
@@ -1,0 +1,29 @@
+﻿using System;
+
+namespace CanHazFunny;
+
+public class Jester
+{
+    public Jester(IJokeOutput jokeOutput, IJokeService jokeService)
+    {
+        JokeOutput = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
+        JokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
+    }
+
+    public IJokeOutput JokeOutput { get; }
+    public IJokeService JokeService { get; }
+
+    public string? Joke { get; set; }
+
+    public void TellJoke()
+    {
+        Joke = JokeService.GetJoke();
+
+        while(Joke.Contains("Chuck Norris", StringComparison.OrdinalIgnoreCase))
+        {
+            Joke = JokeService.GetJoke();
+        }
+
+        JokeOutput.TellJoke(Joke);
+    }
+}
--- a/CanHazFunny/CanHazFunny/JokeService.cs
+++ b/CanHazFunny/CanHazFunny/JokeService.cs
@@ -2,14 +2,23 @@
 
 namespace CanHazFunny
 {
-    public class JokeService
+    public class JokeService : IJokeService
     {
         private HttpClient HttpClient { get; } = new();
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
-            return joke;
+            Uri uri = new Uri("https://geek-jokes.sameerkumar.website/api?format=json%22");
+            string joke = HttpClient.GetStringAsync(uri).Result;
+            return JsonFormatStrip(joke);
+        }
+
+        private static string JsonFormatStrip(string jsonString)
+        {
+            string strippedString = jsonString.Remove(0, 10);
+            int index = strippedString.Length;
+            strippedString = strippedString.Remove(index - 3, 3);
+            return strippedString;
         }
     }
 }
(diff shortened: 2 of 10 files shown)


jstanley6:
I will review this

FenrisStargazer:
I will review this.

jstanley6 (commented):

FenrisStargazer (commented):
**Assignment**
The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ❌
_(I have no clue how to tell if it's <90% exactly, but it seems like you've tested everything.)_
I would say though that the test to make sure the joke doesn't contain Chuck Norris might be kinda iffy because what if the first joke it spits out just honestly doesn't contain it? Not like it's filtered out but like the API just returns a joke that doesn't contain it on the very first go. Then the filter isn't used and thus not tested... This probably doesn't really hurt anything but it's just something to think about.

**Fundamentals**
_( LangVersion is correct, but TargetFramework should be net 7.0)_
For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌/✔??
_(I'm unsure if the way you did this is proper because "AssertEquals<T>" exactly like that didn't work for me)_

**Extra Credit**
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ❌
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string. ✔



When it runs it's cutting off the beginning several characters (about 9 or so?) of the joke, so it isn't printing properly. This also means some Chuck Norris jokes slip by the filter because some part of the name is cut off and thus not recognized.

randypipkins:
@FenrisStargazer 
"For this assignment, always use Assert.AreEqual() (the generic version) ❌/✔??
(I'm unsure if the way you did this is proper because "AssertEquals< T >" exactly like it's shown didn't work for me)"

Not too sure as to why it is not working for you. Took the code and ran the test again a couple times and all of the tests passed fine, even after updating the target frame work to net 7.0.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
...
+public class Jester
+{
+    public Jester(IJokeOutput jokeOutput, IJokeService jokeService)
+    {
+        JokeOutput = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
+        JokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
+    }
+
+    public IJokeOutput JokeOutput { get; }
+    public IJokeService JokeService { get; }
```
mmwoodfo, line 14:
```suggestion
    private IJokeOutput _JokeOutput;
    private IJokeService _JokeService;
```
If these don't need to be public, make them private. Right now, they are only being used within this class.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    public string? Joke { get; set; }
```
mmwoodfo, line 16: Keep this scoped to the TellJoke method, there's no reason you want to be able to access this outside of that method.

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,14 +2,23 @@
...
-    public class JokeService
+    public class JokeService : IJokeService
     {
         private HttpClient HttpClient { get; } = new();
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
-            return joke;
+            Uri uri = new Uri("https://geek-jokes.sameerkumar.website/api?format=json%22");
```
mmwoodfo, line 11:
```suggestion
            Uri uri = new Uri("https://geek-jokes.sameerkumar.website/api?format=json");
```
Don't include this double quote in your URL, you aren't actually getting the json format back.

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,14 +2,23 @@
...
+            string joke = HttpClient.GetStringAsync(uri).Result;
+            return JsonFormatStrip(joke);
+        }
+
+        private static string JsonFormatStrip(string jsonString)
+        {
+            string strippedString = jsonString.Remove(0, 0);
+            int index = strippedString.Length;
+            strippedString = strippedString.Remove(index, 0);
+            return strippedString;
```
mmwoodfo, line 21: This is just stripping the joke string, not deserializing the JSON object, what you want to do is something like this:
`JokeInformation joke = JsonSerializer.Deserialize<JokeInformation>(json);`
```
public class JokeInformation
{
    [JsonPropertyName("joke")] 
    public string? Joke { get; set; }
}
```

PR #67, opened 2023-01-30
Assignment3 @ Joshua-Pedersen

slanante:
Initial PR. Attempted Extra Credit portion to process a JSON format. Used the Newtonsoft.Json library. You will need to configure your VS or VSCode to handle the library or it will not compile.

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,10 +1,23 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
 
 namespace CanHazFunny.Tests
 {
     [TestClass]
     public class JesterTests
     {
+        private Mock<JokeService>? _mockGetJoke;
+        private Mock<JokeWriter>? _mockWriteJoke;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            _mockGetJoke = new();
+            _mockWriteJoke = new();
+        }
+
+
+
         [TestMethod]
         public void GetJoke_CreateJoke_ReturnsJoke()
         {
@@ -37,6 +50,19 @@ namespace CanHazFunny.Tests
             // Assert
 
             Assert.IsFalse((testJoke.Joke!).Contains("Chuck Norris"));
+        }
+
+        [TestMethod]
+        public void FilterJoke_If_ChuckNorris()
+        {
+            // Arrange
+            Jester jester = new Jester(_mockGetJoke!.Object, _mockWriteJoke!.Object!);
+            _mockGetJoke.SetupSequence(joke => joke.GetJoke()).Returns("Chuck Norris");
+            // Act
+            jester.TellJoke();
+            
+            // Assert
+            _mockWriteJoke.Verify(joke => joke.SayJoke("Chuck Norris"), Times.Never);
         }
     }
 }
--- a/CanHazFunny/CanHazFunny/CanHazFunny.csproj
+++ b/CanHazFunny/CanHazFunny/CanHazFunny.csproj
@@ -4,6 +4,7 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net7.0</TargetFramework>
     <LangVersion>11.0</LangVersion>
+    <Nullable>enable</Nullable>
 	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
 	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
 	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
--- a/CanHazFunny/CanHazFunny/CanHazFunny.csproj
+++ b/CanHazFunny/CanHazFunny/CanHazFunny.csproj
@@ -1,13 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net7.0</TargetFramework>
     <LangVersion>11.0</LangVersion>
     <Nullable>enable</Nullable>
-	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
-	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
   </PropertyGroup>
-
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.2"/>
+  </ItemGroup>
 </Project>
--- a/CanHazFunny/CanHazFunny/CanHazFunny.csproj
+++ b/CanHazFunny/CanHazFunny/CanHazFunny.csproj
@@ -2,7 +2,11 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
+    <LangVersion>11.0</LangVersion>
+	  <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
   </PropertyGroup>
 
 </Project>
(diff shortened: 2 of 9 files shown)


CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,15 +1,22 @@
...
         private HttpClient HttpClient { get; } = new();
 
+        public string? Joke { get; set; }
+
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
-            return joke;
+            Joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
+            dynamic? obj = JsonConvert.DeserializeObject(Joke);
```
jdriscoll1, line 15: Why are you deserializing the json, then reserializing it, then deserializing it again manually? Once it's deserialized you can parse the information, or instead manually deserialize the initially received json without deserializing it.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Diagnostics.Contracts;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private JokeService _jokeService;
+        private JokeWriter _jokeWriter;
+        
+        public Jester(JokeService getJoke, JokeWriter writeJoke)
```
jdriscoll1, line 11: Objects should be named as a noun, not as a verb for easier readability. A jokeGetter gets a joke sounds better than A getJoke gets a joke.

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -1,8 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
+    <LangVersion>11.0</LangVersion>
```
jdriscoll1, line 5: Change to latest release

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -5,6 +5,38 @@ namespace CanHazFunny.Tests
...
+            JokeWriter testWriter = new();
+            Jester testJester = new Jester(testJoke, testWriter);
+
+            // Act
+
+            testJester.TellJoke();
+            
+            // Assert
+
+            Assert.IsFalse((testJoke.Joke!).Contains("Chuck Norris"));
```
jdriscoll1, line 39: Testing one joke to see if it contains Chuck Norris is not thorough enough, either test many many jokes, or manually create a test that returns a Chuck Norris joke and validate that it does not go through.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -5,6 +5,38 @@ namespace CanHazFunny.Tests
...
+
+            string generatedJoke = testJoke.GetJoke();
+
+            // Assert
+
+            Assert.IsNotNull(generatedJoke);
+        }
+
+        [TestMethod]
+        public void TellJoke_GenAndWrite_DoesNotContainChuck()
```
jdriscoll1, line 25: Avoid abridging Generate

jdriscoll1 (commented):
Testing if one joke contains Chuck Norris is not thorough enough. It is better to implement another method that makes sure the joke is Chuck Norris and then makes sure it does not go through

Fundamentals
Be sure you enable:
Incorrectly set to 11.0, set to latest instead
Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ❌ 
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string. ✔

CanHazFunny/CanHazFunny/CanHazFunny.csproj
slanante: Isn't net7.0 the latest release? The snippet of code attached on your review shows that net7.0 was added.

CanHazFunny/CanHazFunny/Jester.cs
slanante: Thanks! I'll go ahead and change the variable names.

CanHazFunny/CanHazFunny/JokeService.cs
slanante: Yeah, that was messy. I rewrote this after getting a better understanding of how Json objects work.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
slanante: Great idea, will get that added!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,68 @@
...
+            // Assert
+
+            Assert.IsFalse((testJoke.Joke!).Contains("Chuck Norris"));
+        }
+
+        [TestMethod]
+        public void FilterJoke_If_ChuckNorris()
+        {
+            // Arrange
+            Jester jester = new Jester(_mockGetJoke!.Object, _mockWriteJoke!.Object!);
```
JLee35, line 59: `Object!` unnecessary.
```suggestion
            Jester jester = new Jester(_mockGetJoke!.Object, _mockWriteJoke!.Object);
```

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,68 @@
...
+using Moq;
 
 namespace CanHazFunny.Tests
 {
     [TestClass]
     public class JesterTests
     {
-        
+        private Mock<JokeService>? _mockGetJoke;
+        private Mock<JokeWriter>? _mockWriteJoke;
```
JLee35, line 10:
```suggestion
        private Mock<IGetJoke>? _mockGetJoke;
        private Mock<IWriteJoke>? _mockWriteJoke;
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Diagnostics.Contracts;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private JokeService _jokeService;
+        private JokeWriter _jokeWriter;
```
JLee35, line 9: The point of using interfaces is to add a layer of abstraction between the depending class and the dependency. If we set the types of these two members to a concrete class, we would have to update this code if we wanted to use different classes that implemented the same interfaces.
```suggestion
        private IGetJoke_jokeService;
        private IWriteJoke_jokeWriter;
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public Jester(JokeService jokeGetter, JokeWriter jokeWriter)
```
JLee35, line 11:
```suggestion
        public Jester(IGetJoke jokeGetter, IWriteJoke jokeWriter)
```

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,15 +1,20 @@
...
-        public string GetJoke()
+        public string? Joke { get; set; }
+
+        public virtual string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
-            return joke;
+            Joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
+            dynamic? obj = JsonConvert.DeserializeObject(Joke);
+            string json = obj!.joke;
```
JLee35, line 16: Using `dynamic?` tells the compiler that `JsonConvert.DeserializeObject(Joke)` might return a `null` reference, but then on line 16 you are referencing `obj!.joke` telling the compiler you are sure `obj` is not null. How do you know `obj` is not null if you haven't checked it? What if the user's internet connection dropped before making the API request?
```suggestion
            dynamic? obj = JsonConvert.DeserializeObject(Joke);
            
            if (obj is null) 
            {
               // Either try again, or let the user know a problem occurred.
            }
            string json = obj!.joke;
```

CanHazFunny/CanHazFunny/JokeWriter.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace CanHazFunny
+{
+    public class JokeWriter : IWriteJoke
+    {
+
+        public virtual void SayJoke(string joke)
```
JLee35, line 8: `virtual` wouldn't be required here if you had used the interfaces when setting up your Mock.

JLee35:
Good job on the extra credit, but the main point of using interfaces when declaring dependencies was missed. Read up on the Dependency Inversion Principal [here.](https://en.wikipedia.org/wiki/SOLID)

PR #68, opened 2023-01-30
Assignment3: jdriscoll1 and alowe4

jdriscoll1:


Afterlyte:
I will review this

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,89 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
 
+[assembly: CLSCompliant(true)]
```
Afterlyte, line 5: I'm not sure if this is necessary since part of the requirements for all of us is to use the same (latest) version of the language

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny
+{
+    public interface IJokeOutput
+    {
+        public void Output(string joke);
```
Afterlyte, line 11: I don't think the public modifier should be here since 1) it's public by default, and 2) the point is that it needs to be overridden regardless so making it optional is pointless. (applies to IJokeService too)

CanHazFunny/CanHazFunny/IJokeService.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+[assembly: CLSCompliant(true)]
+namespace CanHazFunny
+{
+    public interface IJokeService
+    {
+        public string GetJoke();
```
Afterlyte, line 12: Same as IJokeOutput with the public modifier

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
...
+        private readonly JokeOutput _JokeOutput;
+        private readonly JokeService _JokeService; 
+        public Jester(JokeOutput jokeOutput, JokeService jokeService) {
+            _JokeOutput = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
+            _JokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
+        
+        }
+
+        public static bool ContainsChuckNorris(string joke) {
+            return joke?.IndexOf("Chuck Norris", StringComparison.OrdinalIgnoreCase) > -1;
```
Afterlyte, line 22: Interesting approach to this, but using .Contains() might be a simpler and more readable approach
```suggestion
            return joke?.Contains("Chuck Norris"); 
```

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,15 +1,23 @@
-﻿using System.Net.Http;
+﻿using System;
+using System.Net.Http;
+using System.Text.Json;
+using static System.Net.Mime.MediaTypeNames;
 
-namespace CanHazFunny
+#pragma warning disable CA1014
+// The code that's violating the rule is on this line.
```
Afterlyte, line 7: I'm not sure if this was some sort of placeholder but I commented this warning disable out and it built without errors

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private readonly JokeOutput _JokeOutput;
+        private readonly JokeService _JokeService; 
+        public Jester(JokeOutput jokeOutput, JokeService jokeService) {
+            _JokeOutput = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
+            _JokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
```
Afterlyte, line 17: This null checking should probably be done in wrapping properties instead of having just readonly fields. That way they can be referenced as properties in other parts of this class so the null checks will apply there too.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+        public string GetJoke() {
+            string joke;
+            do
+            {
+                joke = _JokeService.GetJoke();
+
+            } while (ContainsChuckNorris(joke)); 
+            return joke; 
+        }
```
Afterlyte, line 32: Specifications stated to do the joke retrieval from the API and string checking in the TellJoke method, so this functionality needs to be moved there

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,89 @@
...
+        {
+            // Arrange
+            Jester jester = new Jester(new JokeOutput(), new JokeService());
+
+            // Act
+            string joke = jester.GetJoke();
+            // Assert
+            Assert.IsNotNull(joke);
+            
+        }
```
Afterlyte, line 36: A test should also be written, or this test should be modified, to actually test if the method is returning something that's valid and correct, not just whether or not it's null. Otherwise your code could theoretically be modifying the string from the API before returning it and this still would pass. This is also why the requirement for using generics in the Assert statements is there.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,89 @@
...
+            using StringWriter stringWriter = new();
+            Console.SetOut(stringWriter);
+
+            // Act
+            jester.TellJoke();
+
+            // Assert
+            Assert.IsNotNull(stringWriter.ToString());
+
+        }
```
Afterlyte, line 86: This should also test if the string in the stringwriter is valid, rather than just not null, but the functionality is at least implied here.

Afterlyte (commented):
- Create *two* interfaces. One to represent the `JokeService` and one interface to represent outputting the joke to the screen (`Console.WriteLine`). ✔
- The `JokeService` will need to have the interface applied to it. ✔
- Create an implementation of the output interface that writes out the joke to the console. ✔
- Implement the `Jester` class. It should take in both interfaces as dependencies. These dependencies should be null checked. ❌
  - Change null checking to be done in wrapping properties instead of constructor
- The `Jester` class `TellJoke()` method should retrieve a joke from the `JokeService`. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ❌
  - Move joke retrieval and checking to TellJoke method
- Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔
  - A couple lines are only partially covered, probably because it checked successful running and not exceptions being thrown (like in the constructor) so this could be improved for better coverage. The tests on the strings returned should also be improved to check the strings more accurately.

- Be sure you enable:
  - Nullability for all projects  ✔
  - Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional) ❌
    - TargetFramework needs to be set to net7.0
  - Goes along with comment about the tests

- Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔
- The Geek jokes API that is being used can also return jokes in a JSON format. Update the `JokeService` to retrieve jokes using JSON. `GetJoke` should still return a string. ✔
   - I don't know enough about using the JSON format, but given that JSON functions were used, the code runs properly and tests pass I'd say it's safe to assume it was done correctly, good job!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
jdriscoll1: Thank you for your input. I really appreciate your thoroughness, my partner and I went through and fixed everything up. If you look at the commits you can see precisely what we fixed. Thanks again.

CanHazFunny/.editorconfig
```diff
@@ -0,0 +1,84 @@
+﻿[*.cs]
+
+# CA2234: Pass system uri objects instead of strings
```
JLee35, line 3: Next time add `.editorconfig` to `.gitignore`.

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,8 +1,13 @@
...
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
-
+    <TargetFramework>net7.0</TargetFramework>
+    <EnableNETAnalyzers>True</EnableNETAnalyzers>
+    <EnforceCodeStyleInBuild>True</EnforceCodeStyleInBuild>
+    <AnalysisLevel>5.0-recommended</AnalysisLevel>
+    <Nullable>enable</Nullable>
+    <LangVersion>preview</LangVersion>
```
JLee35, line 9: To keep things simple, while in this class set this to `11.0`.

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,8 +1,13 @@
...
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
```
JLee35, line 10: Next time just use `TreatWarningsAsErrors`.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
JLee35: As @Afterlyte addressed, using `<LangVersion>11.0</LangVersion>` would have prevented you from needing this.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,102 @@
...
+        {
+            // Arrange
+            Jester jester = new(new JokeOutput(), new JokeService());
+
+            // Assert
+            Assert.IsInstanceOfType(jester, typeof(Jester));   
+            
+        
+
+
```
JLee35, line 23: Make a habit of cleaning up blank spaces like this. We want our code to be orderly and clean.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,102 @@
...
+        }
+
+        [TestMethod]
+        public void TestTellJokeReturnTrueIfStringReturned()
```
JLee35, line 27: To follow the test naming guidelines and make the names more readable, consider using underscores to separate the three parts to the name (whatYouAreTesting_WithWhatCondition_ExpectedResult)

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -2,7 +2,17 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
+    <EnableNETAnalyzers>True</EnableNETAnalyzers>
+    <EnforceCodeStyleInBuild>True</EnforceCodeStyleInBuild>
+    <AnalysisLevel>5.0-recommended</AnalysisLevel>
+    <Nullable>enable</Nullable>
+    <LangVersion>preview</LangVersion>
```
JLee35, line 10:
```suggestion
    <LangVersion>11.0</LangVersion>
```

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -2,7 +2,17 @@
...
+    <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
```
JLee35, line 11:
```suggestion
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                ArgumentNullException.ThrowIfNull(value);
+                _JokeOutput = value;
+
+            }
+
+        }
+        public Jester(IJokeOutput jokeOutput, IJokeService jokeService) {
+            JokeOutput = jokeOutput; 
+            JokeService = jokeService;
```
JLee35, line 43: Part of the requirements were to do null checks here.

JLee35 (commented):
Overall good job, especially on the extra credit. Make sure you both do PR reviews so you don't miss out on 20 points.

CanHazFunny/CanHazFunny/Jester.cs
JLee35: Disregard, I missed your null check above.

JLee35:
Disregard my last comment on the PR reviews, I was able to find both your PR reviews. My apologies.

PR #69, opened 2023-01-30
Assignment3

AnnikaStachowiak:
Partnered with @DaisyPalencia
Initial submission

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,10 +1,30 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
 
 namespace CanHazFunny.Tests
 {
     [TestClass]
     public class JesterTests
     {
-        
+        [TestMethod]
+        public void TellJoke_ReturnsJoke_Success()
+        {
+            string tmp = "Joke";
+            Mock<IJokeService> mock = new Mock<IJokeService>();
+            mock.Setup(jokeService => jokeService.GetJoke()).Returns("joke");
+
+            Assert.AreEqual<string>(tmp, mock.Object.GetJoke());
+         }
+        [TestMethod]
+        public void TellJoke_DoesNotReturnChuckNorrisJoke_Success()
+        {
+            JokeService testJoke = new();
+            JokeWriter testWriter = new();
+            Jester testJester = new Jester(testJoke, testWriter);
+
+            testJester.TellJoke();
+
+            Assert.IsFalse((testJoke.Joke!).Contains("Chuck Norris"));
+        }
     }
 }
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -9,22 +9,34 @@ namespace CanHazFunny.Tests
         [TestMethod]
         public void TellJoke_ReturnsJoke_Success()
         {
-            string tmp = "Joke";
+            string tmp = "joke";
             Mock<IJokeService> mock = new Mock<IJokeService>();
             mock.Setup(jokeService => jokeService.GetJoke()).Returns("joke");
 
             Assert.AreEqual<string>(tmp, mock.Object.GetJoke());
          }
         [TestMethod]
-        public void TellJoke_DoesNotReturnChuckNorrisJoke_Success()
+        public void TellJoke_DoesNotReturnChuckNorrisJokeUpper_Success()
         {
             JokeService testJoke = new();
-            JokeWriter testWriter = new();
-            Jester testJester = new Jester(testJoke, testWriter);
+            JokeOutput testOutput = new();
+            Jester testJester = new Jester(testJoke, testOutput);
 
             testJester.TellJoke();
 
-            Assert.IsFalse((testJoke.Joke!).Contains("Chuck Norris"));
+            Assert.IsFalse(condition: !(!testOutput.ToString().ToUpper().Contains("CHUCK") || !testOutput.ToString().ToUpper().Contains("NORRIS")));
+        }
+
+        [TestMethod]
+        public void TellJoke_DoesNotReturnNull_Success()
+        {
+            JokeService testJoke = new();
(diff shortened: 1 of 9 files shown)


CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,30 @@
...
+            Mock<IJokeService> mock = new Mock<IJokeService>();
+            mock.Setup(jokeService => jokeService.GetJoke()).Returns("joke");
+
+            Assert.AreEqual<string>(tmp, mock.Object.GetJoke());
+         }
+        [TestMethod]
+        public void TellJoke_DoesNotReturnChuckNorrisJoke_Success()
+        {
+            JokeService testJoke = new();
+            JokeWriter testWriter = new();
```
randypipkins, line 22: Line 22 was causing the test TellJoke_DoesNotReturnChuckNorrisJoke_Success() to fail
"The type or namespace 'JokeWriter' could not be found"
Would recommend changing it from 'JokeWriter' to 'JokeOutput'(?)

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,30 @@
...
+            Jester testJester = new Jester(testJoke, testWriter);
+
+            testJester.TellJoke();
+
+            Assert.IsFalse((testJoke.Joke!).Contains("Chuck Norris"));
```
randypipkins, line 27: Same issue as with Line 22
"'JokeService' does not contain a definition for 'Joke' and no accessible extension method 'Joke' accepting a first argument of type 'JokeService' could be found"
May have meant to use 'joke' rather than 'Joke'? Since 'joke' is defined in the JokeService class.

randypipkins (commented):
Code looks good other than two issues causing a test to fail. I would recommend adding some null checks in your unit test to ensure that that there is a joke being pulled from the url in JokeService and that 'joke' is not null.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
AnnikaStachowiak: I changed it, thanks!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
AnnikaStachowiak: Changed!

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,6 @@
+﻿using System;
+
+public interface IJokeOutput
+{
+	public void TellJoke(string joke);
```
mmwoodfo, line 5:
```suggestion
	void TellJoke(string joke);
```
Don't use access modifiers in your interface

CanHazFunny/CanHazFunny/IJokeService.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace CanHazFunny;
+
+public interface IJokeService
+{
+	public string GetJoke();
```
mmwoodfo, line 5:
```suggestion
	string GetJoke();
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+namespace CanHazFunny;
+
+public class Jester
+{
+	public JokeService _jokeService { get; set; }
+	public JokeOutput _jokeOutput { get; set; }
```
mmwoodfo, line 7:
```suggestion
	private readonly JokeService _JokeService;
	private readonly JokeOutput _JokeOutput;
```
Do these need to be public? And make sure you are using the correct naming conventions

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
...
+
+	public Jester (JokeService jokeService, JokeOutput jokeOutput)
```
mmwoodfo, line 9: Your formatting makes your code hard to read, make sure you are formatting it correctly. Remember, you can always do `ctrl+K,D` to do some auto formatting!

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
...
+		{
+			throw new ArgumentNullException(nameof(_jokeService));
+		}
+		if (jokeOutput is null)
+		{
+            throw new ArgumentNullException(nameof(_jokeOutput));
+        }
+		_jokeService = jokeService;
+		_jokeOutput = jokeOutput;
+	}
```
mmwoodfo, line 21:
```suggestion
	{
	    _jokeService = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
            _jokeOutput = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
	}
```
You can do the null checking inline, also, make sure you are putting the argument inside the null exception, not the thing you are setting to the argument.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
...
+
+	public void TellJoke()
+	{
+	
+        string joke = _jokeService.GetJoke();
+
+                    while (joke.Contains("Chuck") || joke.Contains("Norris") || joke.Contains("chuck") || joke.Contains("norris"))
```
mmwoodfo, line 28:
```suggestion
                    while (joke.ToLower().Contains("chuck norris")) 
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,35 @@
...
+	public void TellJoke()
+	{
+	
+        string joke = _jokeService.GetJoke();
+
+                    while (joke.Contains("Chuck") || joke.Contains("Norris") || joke.Contains("chuck") || joke.Contains("norris")) 
+			{
+				joke = _jokeService.GetJoke();
+			}
+			Console.WriteLine(joke);
```
mmwoodfo, line 32: You should be using the method on your JokeOutput dependency

PR #73, opened 2023-01-31
Assignment3

Bhopss152:
Joshua Hess
J0shh3ss

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,21 +1,60 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using System;
+using Moq;
 
 namespace CanHazFunny.Tests
 {
     [TestClass]
     public class JesterTests
     {
+
         [TestMethod]
-        public void Jester_Has_JokeService()
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Jester_JokeOutput_Throws_Null_Exception_For_JokeOutput()
         {
             //Arrange
-            IJokeService jokeService = new JokeService();
-            IJokeOutput jokeOut = new JokeOutput();
+            IJokeOutput? jokeOutput = null;
             //Act
-            Jester jester = new(jokeService, jokeOut);
+            new Jester(new JokeService(), jokeOutput);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+          public void Jester_IJokeService_ThrowsNullException_For_Joke_Service()
+        {
+            //Arrange
+            IJokeService? jokeService = null;
+            //Act
+            Jester jester = new Jester((JokeService?)jokeService, new JokeOutput());
+          
+        }
+
+        [TestMethod]
+        public void TellJoke_TestsValidity()
+        {
+            //Arrange
+            string temp = "Test Joke";
+            Mock<IJokeService> mock = new();
+            //Act
+            mock.Setup(jokeService => jokeService.GetJoke()).Returns("Test Joke");
             //Assert
-            Assert.AreEqual<IJokeService>(jester.JokeService, jokeService);
+            Assert.AreEqual<string>(temp, mock.Object.GetJoke());
+
+        }
+
+        [TestMethod]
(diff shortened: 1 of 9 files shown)


CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Reflection.Metadata.Ecma335;
+
+namespace CanHazFunny;
+
+public class Jester
+{
+    public IJokeService JokeService { get; }
+    public IJokeOutput JokeOutput { get; }
+    public Jester(JokeService? _jokeService,IJokeOutput? _jokeOutput)
```
alowe4, line 10: Jester needs to take both interfaces as dependencies, this should be IJokeService and not JokeService.
the rest looks great, I noticed no warnings, it compiles and passes all tests.

alowe4 (commented):
    Implement the Jester class. It should take in both interfaces as dependencies. These dependencies should be null checked. ❌

    Be sure you enable:

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,60 @@
...
+            Mock<IJokeService> mock = new();
+            //Act
+            mock.Setup(jokeService => jokeService.GetJoke()).Returns("Test Joke");
+            //Assert
+            Assert.AreEqual<string>(temp, mock.Object.GetJoke());
+
+        }
+
+        [TestMethod]
+        public void TellJoke_Checks_For_ChuckNorris()
```
mmoran314, line 45: This works in cases where the original joke being used in the method contains "Chuck Norris", in which the program would still work and the test will pass. However the test may also pass if the first joke being tested does not contain "Chuck Norris". I would consider rewriting this test with Moq to make sure that when the program does receive "Chuck Norris" it is properly filtered.

mmoran314 (commented):

Looks good to me for the most part. Would redo the test for the check chuck norris to ensure that it still works when given a chuck norris joke. As it is now there is potential that the first string being tested will not contain Chuck Norris and the test will pass.
Be sure you enable:

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,7 @@
+﻿using System;
+namespace CanHazFunny;
+
+public interface IJokeOutput
+{
+    public void Write(string output);
```
mmwoodfo, line 6:
```suggestion
    void Write(string output);
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Reflection.Metadata.Ecma335;
+
+namespace CanHazFunny;
+
+public class Jester
+{
+    public IJokeService JokeService { get; }
+    public IJokeOutput JokeOutput { get; }
```
mmwoodfo, line 9: Do these need to be publically accessible? Maybe make them private.
```suggestion
    private IJokeService _JokeService;
    private IJokeOutput JokeOutput;
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
...
+    public Jester(IJokeService? _jokeService,IJokeOutput? _jokeOutput)
```
mmwoodfo, line 10:
```suggestion
    public Jester(IJokeService jokeService, IJokeOutput jokeOutput)
```
If you are going to throw a null exception, don't mark the parameters as nullable. Marking them as nullable here says you think it's okay to pass null into your constructor.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
...
+	{
+        JokeService = _jokeService ?? throw new ArgumentNullException(nameof(_jokeService));
+        JokeOutput = _jokeOutput ?? throw new ArgumentNullException(nameof(_jokeOutput));
+    }
+    public string? Joke { get; set; }
```
mmwoodfo, line 15: Keep this scoped to the method you're using it in.

PR #75, opened 2023-01-31
Assignment 3 - Ian Kaiserman

Afterlyte:
Paired with @eroadhouse

EC attempt on Write unit test

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,10 +1,61 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using System;
+using System.IO;
 
-namespace CanHazFunny.Tests
+namespace CanHazFunny.Tests;
+
+[TestClass]
+public class JesterTests
 {
-    [TestClass]
-    public class JesterTests
+    [TestMethod]
+    public void TellJoke_GivenJoke_ReturnJoke()
     {
-        
+        // Arrange
+        var jokeServiceMock = new Mock<IJoke>();
+        JokeService jokeService = new();
+        jokeServiceMock.SetupSequence(jokeService => jokeService.GetJoke()).Returns("Test joke");
+        Jester testJester = new(jokeServiceMock.Object, new ConsoleOutput());
+
+        // Act
+        string joke = testJester.TellJoke();
+
+        // Assert
+        Assert.AreEqual<string>("Test joke", joke);
+    }
+
+    [TestMethod]
+    public void TellJoke_GivenChuckNorris_SkipJoke()
+    {
+        // Arrange
+        var jokeServiceMock = new Mock<IJoke>();
+        JokeService jokeService = new();
+        jokeServiceMock.SetupSequence(jokeService => jokeService.GetJoke())
+            .Returns("Chuck Norris")
+            .Returns("Test joke");
+        Jester testJester = new(jokeServiceMock.Object, new ConsoleOutput());
+
+        // Act
+        string joke = testJester.TellJoke();
+
+        // Assert
+        Assert.AreEqual<string>("Test joke", joke);
+    }
+
+    [TestMethod]
+    public void ConsoleOutput_Write_WritesLine()
+    {
+        // Arrange
+        string text = "test string";
+        ConsoleOutput consoleOutput = new();
+        StringWriter writer = new();
+        Console.SetOut(writer);
+
+        // Act
+        consoleOutput.Write(text);
+        string result = writer.ToString().Trim();
+
+        // Assert
+        Assert.AreEqual<string>(text, result);
     }
 }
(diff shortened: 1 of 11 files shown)


zlawson-ewu:
I will review this.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,61 @@
...
+    }
+
+    [TestMethod]
+    public void ConsoleOutput_Write_WritesLine()
+    {
+        // Arrange
+        string text = "test string";
+        ConsoleOutput consoleOutput = new();
+        StringWriter writer = new();
+        Console.SetOut(writer);
```
zlawson-ewu, line 52: I like your method of testing this. I wish I'd known about StringWriter before I thought to use a file.

zlawson-ewu (commented):
Assignment:

Extra Credit:

  Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔
  
  The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string. ❌

Comments:
  Overall good work! All tests passed, no warnings. Running Main() works as intended. The ConsoleOutput class is tested in a nice and simple way. The code coverage is 100% for Jester and ConsoleOutput, though I don't think the jester tests are quite right for reasons mentioned in the line comment. If the Verify() method is used in the tests instead, I'd consider that bullet point met.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    public string TellJoke()
+    {
+        string joke = Joke.GetJoke();
+        while (joke.Contains("Chuck Norris"))
+        {
+            joke = Joke.GetJoke();
+        }
```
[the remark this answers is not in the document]
Afterlyte: This is a good idea, but when I went to implement it I noticed I still need the extra line to create the variable, otherwise the string obtained in the do/while loop isn't accessible to the rest of the method.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
Afterlyte: Thank you! I remember doing something similar in Java so I did a little researched and found something similar can be done here.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,61 @@
...
+            .Returns("Chuck Norris")
+            .Returns("Test joke");
+        Jester testJester = new(jokeServiceMock.Object, new ConsoleOutput());
+
+        // Act
+        string joke = testJester.TellJoke();
+
+        // Assert
+        Assert.AreEqual<string>("Test joke", joke);
+    }
```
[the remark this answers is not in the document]
Afterlyte: Interesting. we'll look into it a bit, but I'm not sure how necessary it is, specifically since we also did the extra credit block of unit testing the write method on its own. Because realistically if the string is being returned it means it has to have gotten through that line of code successfully, otherwise it would throw an exception of sorts, something else would fail, etc. The main purpose of having TellJoke return a string was for testing purposes, it's the most straightforward way we could think of being able to test input/output using Moq in this case. 

Long story short I'm pretty confident in the fact that it's marked as 100% code coverage and doesn't count it as "partial coverage" or anything like that, but we'll look into getting Verifiable working

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -1,8 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
-  <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
-  </PropertyGroup>
-
+  <ItemGroup>
```
JLee35, line 2: Next time remember lang version, target framework, nullability, NET analyzers, and treat warnings as errors.

CanHazFunny/CanHazFunny/IOutput.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IOutput
+{
+    void Write(String text);
```
JLee35, line 11:
```suggestion
    void Write(string text);
```

CanHazFunny/Directory.Build.props
```diff
@@ -0,0 +1,14 @@
...
+	<PropertyGroup>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net7.0</TargetFramework>
+		<LangVersion>11.0</LangVersion>
+		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+		<WarningsAsErrors />
+		<EnableNETAnalyzers>true</EnableNETAnalyzers>
+		<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
```
JLee35, line 12: Nice!

JLee35:
Great job!

PR #74, opened 2023-01-31
Assignment 3 with Vera Korchemnaya (@VeraKorchemnaya)

Wvvern:
Extra credit was attempted

PR #76, opened 2023-01-31
Assignment3 1bbartels

1bbartels:


VeraKorchemnaya:
I will review

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,82 @@
...
+            }
+            catch (ArgumentNullException e)
+            {
+                expectedException = e;
+                throw e;
+            }
+
+            //Assert
+            Assert.AreEqual(typeof(ArgumentNullException), expectedException);
+        }
```
VeraKorchemnaya, line 29: Excluding this statement on top: [ExpectedException(typeof(ArgumentNullException))] this is the old way of doing it. In class, the professor showed us that we can just use the data row to catch the exception. That would mean we can clean this up significantly: 
```suggestion
        [TestMethod]
        [ExpectedException(typeof(ArgumentNullException))]
        public void Jester_WithNullJokeService_ThrowsException()
        {
                new Jester(null, new JokeOutputter());
        }
```
Because we have the data row it catches the exception for us. If the exception is what we were expecting, the test passes.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,82 @@
...
+
+
+        [TestMethod]
+        //[ExpectedException(typeof(ArgumentNullException))]
```
VeraKorchemnaya, line 56: Remove commented code.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,82 @@
...
+            //Arrange
+            TestJokeOutputter testJokeOutputter = new();
+            Jester jester = new Jester(new JokeService(), testJokeOutputter);
+
+            //Act
+            jester.TellJoke();
+
+            //Assert
+            Assert.IsFalse(testJokeOutputter.joke.Contains("Chuck Norris"));
+        }
```
VeraKorchemnaya, line 68: I see now why you set joke equal to "Chuck Norris" in the TellJoke() method. I would recommend redoing this though. You can use Moq to mock an IJokeService and then set up the TellJoke method to either return one thing, or a sequence of things. I think this is what was recommended in the assignment description. 

Take a look at our pull request for an example: 
[Our pull request](https://github.com/IntelliTect-Samples/EWU-CSCD371-2023-Winter/pull/74/files#diff-a61f21906b512d3c3549a2c7d511f23208988330bab922bfdf329e033e3c6a94)

Specifically, take a look at the last two test methods in the JesterTests class because they use both Mock.Setup and Mock.SetupSequence.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,82 @@
...
+
+
+        [TestMethod]
+        //[ExpectedException(typeof(ArgumentNullException))]
+        public void TellJoke_DoesNotContainChuckNorris()
+        {
+            //Arrange
+            TestJokeOutputter testJokeOutputter = new();
+            Jester jester = new Jester(new JokeService(), testJokeOutputter);
```
VeraKorchemnaya, line 61:
```suggestion
            Jester jester = new (new JokeService(), testJokeOutputter);
```

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -1,7 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
```
VeraKorchemnaya, line 5: Please fix formatting here.

VeraKorchemnaya (commented):
Create an implementation of the output interface that writes out the joke to the console. ✔ (Put implementation of method into class not in interface)
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔ (I would recommend trying Moq because it is a powerful tool you might use on other assignments too)
Fundamentals
Be sure you enable:
For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌ You did not have any tests that used this.
Ensure there are no errors or warnings (including code analysis warnings) ✔ No errors or warnings but there were two messages with minor syntax tweaks, so I would take a look at that. 
To see the error list and any messages, click on the view tab and click "Error List" and then build your project CRL+B. 
All of the above should be unit tested ✔ I would recommend a couple more tests in general
Choose simplicity over complexity ❌ (I recommended a couple things for your tests)
Overall good job!!

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                throw new ArgumentNullException(nameof(jokeOutput));
+            }
+                _jokeService = jokeService;
+                _jokeOutputter = jokeOutput;
+        }
+
+        public void TellJoke()
+        {
+            string joke = "Chuck Norris";
```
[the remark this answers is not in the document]
1bbartels: The assignment to make the string "Chuck Norris" was done to make the following while loop execute, but I'll change it to your suggestion anyway because that also works.

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -2,7 +2,10 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+	<EnableNETAnalyzers>true</EnableNETAnalyzers>
+	<LangVersion>11.0</LangVersion>
+	<Nullable>enable</Nullable>
+    <TargetFramework>net7.0</TargetFramework>
```
JLee35, line 8: Don't forget TreatWarningsAsErrors.

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
JLee35: @VeraKorchemnaya great comments, but next time be sure to clone the code and try to run it on your own machine. This code doesn't do anything because line 8 in `Program.cs` wasn't uncommented or adapted to use the newly built classes:

image

image

JLee35:
Overall good job. I had to deduct some points though, because you didn't uncomment line 8 in `Program.cs` and add your own classes to the constructor. So technically, this code did nothing for the user.

image

image

PR #77, opened 2023-01-31
Assignment 3 Michael Moran (mmoran314) and Ben Fristad (BenFristadSchool)

mmoran314:
Filtered Chuck Norris Jokes,
Added Unit Tests
Added JokeOutput Class

### DIFF
--- a/CanHazFunny/CanHazFunny/IJokeOutput.cs
+++ b/CanHazFunny/CanHazFunny/IJokeOutput.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IJokeOutput
+{
+    public void WriteJoke(string joke);
+    
+}
--- a/CanHazFunny/CanHazFunny/IJokeService.cs
+++ b/CanHazFunny/CanHazFunny/IJokeService.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IJokeService
+{
+    public string GetJoke();
+}
--- a/CanHazFunny/CanHazFunny/Jester.cs
+++ b/CanHazFunny/CanHazFunny/Jester.cs
@@ -41,7 +41,7 @@ public class Jester : IJokeOutput, IJokeService
     }
 
     
-    public void TellJoke()
+    public string TellJoke()
     {
         string testJoke = GetJoke();
         bool isChuckNorris = true;
@@ -57,6 +57,7 @@ public class Jester : IJokeOutput, IJokeService
         }
 
         ReallyCoolJokeOutput.WriteJoke(testJoke);
+        return testJoke;
     }
     public string GetJoke()
     {
--- a/CanHazFunny/CanHazFunny/Jester.cs
+++ b/CanHazFunny/CanHazFunny/Jester.cs
@@ -1,0 +1,69 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public class Jester : IJokeOutput, IJokeService
+{
+    private IJokeOutput? _ReallyCoolJokeOutput;
+    private IJokeService? _JokeService;
+    
+    public IJokeOutput ReallyCoolJokeOutput
+    {
+        get
+        {
+            return _ReallyCoolJokeOutput!; // Use null forgiveness here after null checking on setter
+        }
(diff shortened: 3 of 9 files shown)


Nolan-Posey (commented):
Assignment
For this assignment do the following:

Fundamentals
Be sure you enable:

Overall you guys did a great job! The program compiles and runs as intended. There are only two things you missed:

1.)  The code coverage for the Jester class is under 90% (The Fine Code Coverage extension I'm using is say that it is at 55.5%). I would recommend making some mock classes for JokeOutput and JokeService inside of your test class to use inside of a test jester that way you have more control over the tests. Also I would recommend splitting the CreateJester_Failure() test into two tests, one that expects a null argument exception when  JokeOutput is null and another when Joke service is null.

2.) In the TestWriteJokeConsoleOutput() test you guys didn't use the generics version of Assert.AreEqual (Assert.AreEqual<T>). I would recommend using Assert.AreEqual<string> since it appears you are trying to compare the string intended to be output and the string that the Jester class is outputting.

jacobnewcomb:
Fundamentals
Be sure you enable:
For this assignment, always use Assert.AreEqual() (the generic version) ❌

Didn't seem to use generic version of Assert.AreEqual(), but everything else looks great and seems to work well!

mmoran314:
>
> Didn't seem to use generic version of Assert.AreEqual(), but everything else looks great and seems to work well!

added <string> to tests.

mmoran314:
> 1.) The code coverage for the Jester class is under 90% (The Fine Code Coverage extension I'm using is saying that coverage for the Jester class is at 55.5%). I would recommend making some mock classes for JokeOutput and JokeService inside of your test class to use inside of a test jester that way you have more control over the tests. Also I would recommend splitting the CreateJester_Failure() test into two tests, one that expects a null argument exception when JokeOutput is null and another when Joke service is null.
> 
> 2.) In the TestWriteJokeConsoleOutput() test you guys didn't use the generics version of Assert.AreEqual (Assert.AreEqual-T). I would recommend using Assert.AreEqual-string since it appears you are trying to compare the string intended to be output and the string that the Jester class is outputting. (The "T" and "string" should be in angled brackets but for some reason they disappear when I post my comment)

Added generic to tests and implemented mock for properly testing the Chuck Norris filter.
Split the null test into two tests, hopefully covers enough.

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IJokeOutput
+{
+    public void WriteJoke(string joke);
```
mmwoodfo, line 12:
```suggestion
    void WriteJoke(string joke);
```

CanHazFunny/CanHazFunny/IJokeService.cs
```diff
@@ -0,0 +1,12 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IJokeService
+{
+    public string GetJoke();
```
mmwoodfo, line 11:
```suggestion
    string GetJoke();
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public class Jester : IJokeOutput, IJokeService
```
mmwoodfo, line 9:
```suggestion
public class Jester : IJokeOutput, IJokeService
```
these should be implemented on their own classes, not here.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,70 @@
...
+        JokeService = jokeService;
+    }
+
+    
+    public string TellJoke()
+    {
+        string testJoke = GetJoke();
+        bool isChuckNorris = true;
+        while (isChuckNorris) {
+            if (testJoke.Contains("Chuck Norris") || testJoke.Contains("Chuck") || testJoke.Contains("Norris")) {
```
mmwoodfo, line 49: I'd suggest moving this if check into your while loop and getting rid of your `isChuckNorris` boolean all together.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,70 @@
...
+                continue;
+            }
+            else
+            {
+                isChuckNorris=false;
+            }
+        }
+
+        ReallyCoolJokeOutput.WriteJoke(testJoke);
+        return testJoke;
```
mmwoodfo, line 60: You don't need to return a string here, you could make this method void. Just depends what the intention is, I'm guessing you did this to make unit testing easier.

PR #80, opened 2023-01-31
Assignment 3 - Hunter Thompson & Hunter Allen

StrangeRanger:
@hallen10

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -6,17 +6,29 @@ namespace CanHazFunny.Tests;
 [TestClass]
 public class JesterTests
 {
+    private Jester? Jester { get; set; }
+    
+    [TestInitialize]
+    public  void TestInitialize()
+    {
+        TestCleanup();
+        Jester = new Jester();
+    }
+    
+    [TestCleanup]
+    public void TestCleanup()
+    {
+        Jester = null;
+    }
+    
     [TestMethod]
     public void TellJoke_ShouldNotContainChuckNorris()
     {
         StringWriter stringWriter = new();
         Console.SetOut(stringWriter);
 
-        // Arrange
-        Jester jester = new();
-
         // Act
-        jester.TellJoke();
+        Jester.TellJoke();
 
         // Assert
         Assert.IsFalse(stringWriter.ToString().Contains("Chuck Norris"));
@@ -25,13 +37,16 @@ public class JesterTests
     [TestMethod]
     public void TellJoke_ShouldNotReturnNull()
     {
-        // Arrange
-        Jester jester = new();
-
         // Act
-        String joke = jester.GetJoke();
+        string joke = Jester.GetJoke();
 
         // Assert
         Assert.IsNotNull(joke);
     }
+
+    [TestMethod]
+    public void Jester_Initialize_IsNotNull() {
+        Assert.IsNotNull(Jester);
+    }
+
 }
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -7,46 +7,43 @@ namespace CanHazFunny.Tests;
 public class JesterTests
 {
     private Jester? Jester { get; set; }
-    
+
     [TestInitialize]
-    public  void TestInitialize()
+    public void TestInitialize()
     {
         TestCleanup();
-        Jester = new Jester();
+        Jester = new Jester(new JokeOutput(), new JokeService());
     }
-    
+
     [TestCleanup]
     public void TestCleanup()
     {
         Jester = null;
     }
-    
+
     [TestMethod]
     public void TellJoke_ShouldNotContainChuckNorris()
     {
-        StringWriter stringWriter = new();
+        StringWriter stringWriter = new StringWriter();
         Console.SetOut(stringWriter);
 
         // Act
-        Jester.TellJoke();
+        Jester?.TellJoke();
 
         // Assert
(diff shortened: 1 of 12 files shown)


Howard404:
I will review this

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -1,8 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <Import Project="..\Directory.Build.Props" />
+
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
-
```
Howard404, line 7: Should contain <EnableNETAnalyzers>true</EnableNETAnalyzers> & <Nullable>enable</Nullable>
Also suggest: <WarningsAsErrors>true</WarningsAsErrors> & <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,52 @@
...
+    }
+    
+    [TestMethod]
+    public void TellJoke_ShouldNotContainChuckNorris()
+    {
+        StringWriter stringWriter = new();
+        Console.SetOut(stringWriter);
+
+        // Act
+        Jester.TellJoke();
```
Howard404, line 31: Getting warning of possible dereferencing of type null

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,52 @@
...
+
+        // Assert
+        Assert.IsFalse(stringWriter.ToString().Contains("Chuck Norris"));
+    }
+
+    [TestMethod]
+    public void TellJoke_ShouldNotReturnNull()
+    {
+        // Act
+        string joke = Jester.GetJoke();
```
Howard404, line 41: Same warning of dereferencing null

Howard404 (commented):
Just a couple of warnings. Great work! 

- Create *two* interfaces. One to represent the `JokeService` and one interface to represent outputting the joke to the screen (`Console.WriteLine`). ✔
- The `JokeService` will need to have the interface applied to it. ✔
- Create an implementation of the output interface that writes out the joke to the console. ✔
- Implement the `Jester` class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔
- The `Jester` class `TellJoke()` method should retrieve a joke from the `JokeService`. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
- Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔

## Fundamentals

- Be sure you enable:
  - Nullability for all projects  ❌✔
  - Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ❌✔
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  ❌✔
- **Ensure there are no errors or warnings (including code analysis warnings)** ❌✔

CanHazFunny/CanHazFunny/CanHazFunny.csproj
Howard404: Should contain <EnableNETAnalyzers>true</EnableNETAnalyzers> & <Nullable>enable</Nullable>
Also suggest: <WarningsAsErrors>true</WarningsAsErrors> & <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>

StrangeRanger: I specify all those kinds configurations inside the `Directory.Build.Props` file.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
StrangeRanger: Fixed. Thank you.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        get
+        {
+            return _JokeOutput!;
+        }
+        set
+        {
+            _JokeOutput = value ?? throw new ArgumentNullException(nameof(value));
+        }
+    }
```
JLee35, line 17: We really don't want a public setter for a private variable like `_JokeOutput`. The idea is to protect the private variable from being modified from where it shouldn't. So next time, I would encourage either a private setter, or no setter at all and just set the field within the constructor.

CanHazFunny/Directory.Build.Props
```diff
@@ -0,0 +1,11 @@
+<Project>
+
+    <PropertyGroup>
+        <TargetFramework>net7.0</TargetFramework>
+        <LangVersion>11.0</LangVersion>
+        <Nullable>enable</Nullable>
+        <EnableNETAnalyzers>true</EnableNETAnalyzers>
+        <WarningsAsErrors>true</WarningsAsErrors>
```
JLee35, line 8: Nice!

JLee35:
Great job!

CanHazFunny/CanHazFunny/Jester.cs
JLee35: After a discussion with Mark on this, it sounds like this will be addressed in class. There are cases where you would want a public setter for something like this, but my point was that this is not one of those cases. Obviously no points were deducted, you followed the guidelines of the writeup.

StrangeRanger: Alright. Thanks for letting me know. I've had a few questions specifically about the properties and their scope, such as that, so that's perfect.

PR #85, opened 2023-02-05
Assignment4 @ Joshua Pedersen

slanante:
Initial PR


AnnikaStachowiak (commented):
Tasks

Fundamentals

    Be sure you enable:

Extra Credit

Looks good! Though I did find your choice of 'BookName' instead of 'Title' to be interesting. Not an issue, though.

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,7 @@
+namespace Logger;
+
+public abstract class BaseEntity : IEntity
```
mmwoodfo, line 3:
```suggestion
public abstract record BaseEntity : IEntity
```
You'll want to define this as an abstract **_record_** class, that way you can inherit from it on your book employee and student records.

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,7 @@
...
+{
+    public Guid Id { get; init; }
+    public abstract string Name { get; }
```
mmwoodfo, line 6: Good job!

Logger/BookStudentEmpRecords.cs
```diff
@@ -0,0 +1,26 @@
+namespace Logger;
+
+// All three records the IEntity interface are implemented implicitly to ensure
+// we avoid conflicts with members of the same properties
+
+public record Book : IEntity
```
mmwoodfo, line 6: I would reccommend splitting each of these records into their own files.

Logger/Person.cs
```diff
@@ -0,0 +1,10 @@
+namespace Logger;
+
+public record Person : IEntity
+{
+    public Guid Id { get; init; }
+    public FullName FullName { get; init; }
+
+    // Implemented explicitly to avoid conflicts betweeen implementing classes
+    string IEntity.Name => FullName.Name;
```
mmwoodfo, line 9: You mention you are implementing this explicitly, to avoid conflicts, what conflicts would you expect to have? How would you expect Name to be implemented differently for a class inheriting from Person?

It doesn't look like you're impementing this in other classes that's using it.

Logger/BookStudentEmpRecords.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public Guid Id { get; init; }
+    public string BookName { get; init; } = string.Empty;
+    public FullName Author {get; init;}
+    public string ISBN {get; init; } = string.Empty;
```
mmwoodfo, line 11: Consider making these nullable instead?

Logger/BookStudentEmpRecords.cs
```diff
@@ -0,0 +1,26 @@
...
+    // Here Name is calculated
+    public string Name => $"{BookName} , Author: {Author.Name}, ISBN: {ISBN}";
```
mmwoodfo, line 13: Looks good!

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,73 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class EntityTests
+{
+    [TestMethod]
+    public void Students_With_Same_Properties_Are_Equal()
```
mmwoodfo, line 9:
```suggestion
    public void StudentsRecord_CompareTwoIdenticalRecords_AreEqual()
```
Try to name your tests like this:
`MethodBeingTested_TestingScenerio_ExpectedBehavior`

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        Student student2 = new Student
+        {
+            Id = student1.Id,
+            FullName = student1.FullName,
+            StudentYear = student1.StudentYear
+        };
+
+        Assert.AreEqual(student1, student2);
+    }
```
mmwoodfo, line 26:
```suggestion
    {
       // Arrange
        Student student1 = new Student
        {
            Id = Guid.NewGuid(),
            FullName = new FullName("Spongebob", "SquarePants"),
            StudentYear = StudentLevels.Freshman
        };

        Student student2 = new Student
        {
            Id = student1.Id,
            FullName = student1.FullName,
            StudentYear = student1.StudentYear
        };
        
        // Act & Assert
        Assert.AreEqual(student1, student2);
    }
```
Consider using the // Arrange // Act // Assert formatting

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,73 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class EntityTests
```
mmwoodfo, line 6: You may want to write tests to make sure the name property is returning what you'd expect.

PR #86, opened 2023-02-06
Assignment4: jdriscoll1 and alowe4

jdriscoll1:
Extra Credit Attempted


Nolan-Posey (commented):
Be sure you enable:

Everything looks great! The only thing I didn't see was some tests for value equality for the Entities Book, Student, and Employee, but I did see your tests for the FullName record. 

My only suggestions would be:
1.) To add a simple value equality test for Book, Student, and Employee.
2.)  This is totally optional, but instead of having to make a test class for Book, Student, and Employee, I would recommend making a "RecordsTests" test class and then adding all your value equality tests to it. You could even move everything from your FullNameTests class over there if you so choose just to get it all into one file.

Overall, good job!
-Nolan

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -1,24 +1,26 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.Collections.Generic;
 
+[assembly: CLSCompliant(true)]
```
mmwoodfo, line 6: Why add this to your tests?

Logger.Tests/FileLoggerTests.cs
```diff
@@ -6,19 +6,19 @@ namespace Logger.Tests;
 public class FileLoggerTests : FileLoggerTestsBase
 {    
     [TestMethod]
-    public void Create_GivenClassAndValidFileName_Success()
+    public void CreateGivenClassAndValidFileNameSuccess()
```
mmwoodfo, line 9: You can keep the underscores in your test names, this is the typical way to name unit tests in c#

Logger.Tests/RecordTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            Employee e1 = new(name, "Programmer", "Walmart");
+            Employee e2 = new(name, "Programmer", "Walmart");
+            Assert.AreNotEqual<Employee>(e1, e2);
+
+
+        }
+
+
+        [TestMethod]
+        public void BooksAreNotEqual()
```
mmwoodfo, line 63: You should make sure to check if equal as well!

Logger/Records.cs
```diff
@@ -0,0 +1,52 @@
...
+    // that the two objects are tied to one object
+
+    /* MUTABLE OR IMMUTABLE EXPLANATION:  */
+    // Should the full name structure be mutable or immuable? Immutable
+    // There are three reasons that the Full Name record should be immutable
+    // Firstly, whenever we change the name, we should be creating a new Full Name Object rather than changing
+    // a previous one for better clarity
+    // Secondly, it should be immuatable so that it does not change a different object elsewhere in the code
+    // Thirdly, hash codes are calculated from the direct data storage, if we chagne the hash code we could lose information
+    public readonly record struct FullName(string FirstName, string LastName, string? MiddleName = null)
```
mmwoodfo, line 25: You should split these records into their own files.

Logger/IEntity.cs
```diff
@@ -1,6 +1,17 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger
 {
+    public interface IEntity
+    {
+        public Guid Id { get; init; }
```
mmwoodfo, line 12:
```suggestion
        Guid Id { get; init; }
```

Logger/IEntity.cs
```diff
@@ -1,6 +1,17 @@
...
+        public string Name { get; }
```
mmwoodfo, line 13:
```suggestion
        string Name { get; }
```

Logger/BaseLogger.cs
```diff
@@ -5,7 +5,7 @@
 public abstract class BaseLogger
 {
     public string LogSource { get; }
-    public BaseLogger(string logSource) => LogSource = string.IsNullOrWhiteSpace(logSource)
+    protected BaseLogger(string logSource) => LogSource = string.IsNullOrWhiteSpace(logSource)
```
mmwoodfo, line 8: Why make this protected?

Logger/AbstractRecords.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger
+{
+ public abstract record class Entity : IEntity
```
mmwoodfo, line 9: Your file name should reflect this class name.

PR #87, opened 2023-02-06
Assignment4 @randypipkins

avanchu253:
@randypipkins

### DIFF
--- a/Logger.Tests/EntityTests.cs
+++ b/Logger.Tests/EntityTests.cs
@@ -1,0 +1,73 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class EntityTests
+{
+    [TestMethod]
+    public void StudentsWithSamePropertiesAreEqual()
+    {
+        Student student1 = new Student
+        {
+            Id = Guid.NewGuid(),
+            FullName = new FullName("Michael", "Jordan"),
+            StudentYear = StudentLevels.Freshman
+        };
+
+        Student student2 = new Student
+        {
+            Id = student1.Id,
+            FullName = student1.FullName,
+            StudentYear = student1.StudentYear
+        };
+
+        Assert.AreEqual(student1, student2);
+    }
+
+    [TestMethod]
+    public void BooksWithSamePropertiesAreEqual()
+    {
+        FullName author = new FullName("Mark", "Michaelis");
+
+        Book book1 = new Book
+        {
+            Id = Guid.NewGuid(),
+            BookName = "Essential C#",
+            Author = author,
+            ISBN = "9780135972267"
+        };
+
+        Book book2 = new Book
+        {
+            Id = book1.Id,
+            BookName = book1.BookName,
+            Author = book1.Author,
+            ISBN = book1.ISBN
+        };
+
+        Assert.AreEqual(book1, book2);
+        Assert.AreEqual(book1.Name, book2.Name);
+    }
+
+    [TestMethod]
+    public void EmployeesWithSamePropertiesAreEqual()
+    {
+        Employee employee1 = new Employee
+        {
+            Id = Guid.NewGuid(),
+            FullName = new FullName("Deion", "Sanders"),
+            Employer = "University of Colorado"
+        };
+
+        Employee employee2 = new Employee
+        {
+            Id = employee1.Id,
+            FullName = employee1.FullName,
+            Employer = employee1.Employer,
+        };
+
+        Assert.AreEqual(employee1, employee2);
+    }
+
+}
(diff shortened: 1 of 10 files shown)


BenFristadSchool (commented):
I will review this

Logger/BookStudentEmployeeRecords.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Logger;
+
+// All three records the IEntity interface are implemented implicitly to ensure
+// we avoid conflicts with members of the same properties
+
+public record Book : IEntity
```
BenFristadSchool, line 6: Why does Book implement IEntity, rather than extend BaseEntity?

Logger/FullName.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Logger;
+public record struct FullName(string FirstName, string LastName, string? MiddleName = null)
+{
+    public string FirstName { get; } = FirstName??throw new ArgumentNullException(nameof(FirstName));
+    public string LastName { get; } = LastName??throw new ArgumentNullException(nameof(LastName));
+    public string? MiddleName { get; } = MiddleName;
+
+    public string Name => $"{FirstName} {MiddleName} {LastName}";
```
BenFristadSchool, line 8: When using string interpolation here, since middle name is optional, wont it created a double space between first and last name when no middle name is provided?

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class StorageTests
+{
+    [TestMethod]
+    public void Adding_Entities()
```
BenFristadSchool, line 9: This is a good test, but you should also create tests for other storage methods such as get and remove

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        Employee employee2 = new Employee
+        {
+            Id = employee1.Id,
+            FullName = employee1.FullName,
+            Employer = employee1.Employer,
+        };
+
+        Assert.AreEqual(employee1, employee2);
+    }
```
BenFristadSchool, line 71: Could add tests for when not equal

Logger/FullName.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Logger;
+public record struct FullName(string FirstName, string LastName, string? MiddleName = null)
```
BenFristadSchool, line 2: could create tests to ensure proper fullname creation

BenFristadSchool (commented):
Well Done!

Logger/FullName.cs
```diff
@@ -0,0 +1,9 @@
...
+{
+    public string FirstName { get; } = FirstName??throw new ArgumentNullException(nameof(FirstName));
```
mmwoodfo, line 4:
```suggestion
    public string FirstName { get; } = FirstName ?? throw new ArgumentNullException(nameof(FirstName));
```
Small whitespace nit

Logger/BookStudentEmployeeRecords.cs
```diff
@@ -0,0 +1,25 @@
...
+    public Guid Id { get; init; }
+    public string BookName { get; init; } = string.Empty;
+    public FullName Author { get; init; }
+    public string ISBN { get; init; } = string.Empty;
+    public string Name => $"{BookName} , Author: {Author.Name}, ISBN: {ISBN}";
+
+}
+
+// Student and Employee are made as derivatives of the Person class
+public record Student : Person
```
mmwoodfo, line 17: These records should be split into their own files

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Logger;
+
+public abstract record class BaseEntity : IEntity
```
mmwoodfo, line 3: You never wrote anything in the IEntity class

Logger/Person.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Logger;
+
+public record Person : BaseEntity
+{
+    public FullName FullName { get; init; }
+
+    // Implemented explicitly to avoid conflicts betweeen implementing classes
+    public override string Name => FullName.Name;
```
mmwoodfo, line 8: What makes this an explicit implementation?

mmwoodfo: Also, it doesn't seem like you are implementing this anywhere else, so what conflicts would you expect to have?

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,73 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class EntityTests
+{
+    [TestMethod]
+    public void StudentsWithSamePropertiesAreEqual()
```
mmwoodfo, line 9: Use the standard MethodName_TestConditions_ExpectedResult naming convention for unit tests. Also consider using //arrange //act //assert

PR #89, opened 2023-02-07
Assignment4

j0shh3ss:
Brendan Hopkins
Bhopss152


zlawson-ewu:
I will review this.

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,26 @@
...
+            Employer = "University of Washington"
+        };
+
+        // Act
+        storage.Add(entity);
+
+        // Assert
+        Assert.IsTrue(storage.Contains(entity));
+    }
+}
```
zlawson-ewu, line 26: Add() and Contains() are tested but not Remove() or Get(). All three record types should be tested against this as well (only employee tested here).

zlawson-ewu (commented):
        Provide a comment on the full name record on why you selected to define a value or a reference type and ❌

        Add an Id property of type Guid that is an init-only setter. ❌
        Add a Name property that is string. ❌

        Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ❌?✔

Unit test the Storage class using various entities to ensure that there are no bugs. ❌

Commentary: 
Compiles, no warnings or errors, and indeed favors simplicity over complexity. GetHashCode() and Equals() should be overloaded for custom implementations (and tested). Missing tests for everything but Storage class. IEntity is missing members. Didn't evaluate "comment on interface method" because there are no methods. Unsure if Instructor duplicated lines accidentally.

Logger/book.cs
```diff
@@ -0,0 +1,22 @@
...
+
+public record Student : Person
+{
+    public GradeLevel StudentGrade { get; init; }
+}
+
+public record Employee : Person
+{
+    public string Employer { get; init; } = string.Empty;
+}
```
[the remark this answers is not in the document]
j0shh3ss: Great suggestion!

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Logger;
+//implemented implicitly to prevent data loss/interference
+public abstract record class BaseEntity : IEntity
+{
+    public Guid Id { get; init; }
+    public abstract string Name { get; }
+}
```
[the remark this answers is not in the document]
j0shh3ss: Thanks!

j0shh3ss:
Thanks for all the feedback! I changed everything you recommended and then some. Enjoying these reviews!

Logger.Tests/StorageTests.cs
j0shh3ss: Updated!

Logger/FullName.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Logger;
```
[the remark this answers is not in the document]
j0shh3ss: Added!

Logger/book.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger;
+//We defined each record implicity to make sure data loss is preveneted
+//And so that elements can be accessed without interference 
+public record Book : IEntity
```
mmwoodfo, line 4:
```suggestion
public record Book : BaseEntity
```
It looks like you did this for your Person record, but make sure you are inheriting from your base entity!

Also, just for consistency, in your Person, Student, and Employee records, you are adding the `class` keyword to the record definition, technically this is done by default. But you should declare your records in a consistent manner and either add that `class` keyword here to be explicit about it or remove it from your other records and let them be class records implicitly.

Logger/book.cs
```diff
@@ -0,0 +1,8 @@
...
+{
+    public Guid Id { get; init; }
+    public string Name { get; init; } = string.Empty;
```
mmwoodfo, line 7: Maybe consider adding some other properties like a book author, title, SSN, etc. And then using the name to return a concatenation of those different fields.

Logger/Employee.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Logger;
+//We defined each record implicity to make sure data loss is preveneted and all data can be accessed
+public record class Employee : Person
+{
+    public string Employer { get; init; } = string.Empty;
```
mmwoodfo, line 5: Consider passing this in via a constructor rather than defaulting to empty. If you intend for it not to always be set, make it nullable.

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            Id = Guid.NewGuid(),
+            FullName = new FullName("Steve", "Jobs"),
+            Employer = "Employer",
+        };
+        //Assert
+        Assert.AreEqual(f, employee.FullName);
+    }
+    [TestMethod]
+    public void Stu_Imp_Entity()
```
mmwoodfo, line 34: Don't use abbreviations.

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,48 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace Logger.Tests;
+[TestClass]
+public class EntityTests
```
mmwoodfo, line 4: You should have some equality tests here

PR #91, opened 2023-02-07
Assignment4 @Clockknight

Malcolm-Shepherd:


Bhopss152 (commented):
Define an abstract base class that implements IEntity - appropriately choosing to implement the interface explicitly or implicitly. ❌
Do not implement the Name property in this abstract class. ❌
Define book, student, and employee records - all with entity capabilities. ❌
Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ✔
Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.) ❌
Be sure you enable:

I like your code, I marked some things down that were there but that could be changed around a little. I would make your base class more easier to find by giving it the name "base" something like that just makes it faster to find. 
For the "Name" method, why did you choose that layout for returning the values.
Good comments where we need to and minimal comments where we dont need to.✔

Logger/FullName.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Logger;
+public record class FullName(string FirstName, string LastName, string? MiddleName = null)
```
DaisyPalencia, line 2: Could use "is null" here.

Logger/FullName.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+/*Reference type as different people can have the same full name.
+  Read only as it is very rare for people to need to change their name.
```
DaisyPalencia, line 5: I recommend using the word mutable/immutable here, so it's clear which task your comment is for.

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,44 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+
+namespace Logger.Tests;
```
DaisyPalencia, line 4: For this test class, I would recommend adding tests that include null values.

DaisyPalencia (commented):
Tasks
Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ❌
Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.❌
Unit test the Storage class using various entities to ensure that there are no bugs. ✔
Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.) ❌
Pay special attention to the CreateLogger methods throughout. ✔
Fundamentals
Be sure you enable:

Good job overall! I did recommend some changes but I'm sure once you make those your project will be good to go.

Logger/FullName.cs
Malcolm-Shepherd: This is assigning MiddleName, not checking for equality. Otherwise I would agree with you.

Logger.Tests/EntityTests.cs
Malcolm-Shepherd: I thought about adding null checks here, but since it didn't specify that we had to and they say "Choose simplicity over complexity" I didn't want to add any complexity that wasn't asked for.

Logger/FullName.cs
Malcolm-Shepherd: Updating comments to use more clear language in regard to the requirements.

Malcolm-Shepherd:
> Define a full name record (first, last, middle) handling optional and null appropriately. ✔ Provide a comment on the full name record on why you selected to define a value or a reference type and ✔ Provide a comment on the full name record on why or why not the type is immutable. ✔ Define an IEntity interface: ✔ Add an Id property of type Guid that is an init-only setter. ✔ Add a Name property that is string. ✔ Define an abstract base class that implements IEntity - appropriately choosing to implement the interface explicitly or implicitly. ❌ Do not implement the Name property in this abstract class. ❌ Do force any derived classes to provide an implementation for Name. ✔ Define book, student, and employee records - all with entity capabilities. ❌ Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ✔ Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ✔ You should consider the relationship between Student and Employee and refactor the common code shared between them. ✔ Given the properties on the full name record, you should consider which entities it makes sense to use it. ✔ Test that the equality behavior on these entities behaves as expected. ✔ Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ✔ Unit test the Storage class using various entities to ensure that there are no bugs. Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.) ❌ Be sure you enable: Enable Nullability for all projects ✔ Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔ Ensure there are no errors or warnings (including code analysis warnings) ✔ All of the above should be unit tested ✔ Choose simplicity over complexity ✔
> 
> I like your code, I marked some things down that were there but that could be changed around a little. I would make your base class more easier to find by giving it the name "base" something like that just makes it faster to find. For the "Name" method, why did you choose that layout for returning the values. Good comments where we need to and minimal comments where we dont need to.✔

Renamed entity to BaseEntity to make it more obvious and descriptive.
I assume what you ask "why did you choose that layout for returning the values?" you are referring to person. I'm not sure what the proper term it is, but it's basically a if-else block in a singe line (condition ? [if true do this] : [else do this]). I used to properly format the full name checking if it included a middle name.
I also didn't see any issues with the logger classes. The closest thing would be the variable "logggerConfiguration) having 3 g's, but that didn't seem like an issue necessarily.

Logger/IEntity.cs
```diff
@@ -3,5 +3,6 @@ public interface IEntity
 {
 
         // Place members here.
```
mmwoodfo, line 5:
```suggestion
```

Logger/Person.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Logger;
+
+public record class Person(FullName PersonName) : BaseEntity
+{
+    /*Implicity implemented as People have full names, so it only makes sense to take in a FullName*/
+    public override string Name {
+        get
+        {
+            return PersonName.MiddleName == null ? PersonName.FirstName + " " + PersonName.LastName: PersonName.FirstName + " " + PersonName.MiddleName + " " + PersonName.LastName;
```
mmwoodfo, line 15: I'd use string interpolation here

E.g.
`$"{PersonName.FirstName} {PersonName.LasstName}"

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,44 @@
...
+
+[TestClass]
```
mmwoodfo, line 6: You should have some equality tests here as well

Logger/FullName.cs
mmwoodfo, line 2: I think you should make this a struct, making it a struct doesn't prevent users from having the same name.

PR #92, opened 2023-02-07
Assignment4 - Ian Kaiserman

Afterlyte:
paired with @eroadhouse

EC attempted for items/properties in a single file (this was in the original submission but I forgot to note it)


StrangeRanger:
Will Review

Malcolm-Shepherd:
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Provide a comment on the full name record on why or why not the type is immutable. ✔
  - You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌
  - Test that the equality behavior on these entities behaves as expected. ✔
  - Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
- Unit test the Storage class using various entities to ensure that there are no bugs. ✔
- Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.)✔

Malcolm-Shepherd:
## Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. 
No comment on book's implementation of name. I also don't agree with your implementation of name. If I called Book.Name, I would expect the title and not the author's name.
## You should consider the relationship between Student and Employee and refactor the common code shared between them. 
Code wasn't refactored. You copied the shared code into a new interface then didn't use the interface or remove the shared code in the original classes.

StrangeRanger (commented):
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Provide a comment on the full name record on why or why not the type is immutable. ✔
    - **DOESN'T SEEM LIKE YOU PROVIDED ANY COMMENTS ON THE INTERFACE MEMBER THEMSELVES**
  - You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌
    - **WAS REFACTORED BUT IT DOESN'T SEEM LIKE IT WAS EVER IMPLIMENTED INTO EITHER  STUDENT OR EMPLOYEE**
  - Test that the equality behavior on these entities behaves as expected. ✔
  - Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
- Unit test the Storage class using various entities to ensure that there are no bugs.

## Fundamentals

- Be sure you enable:

## Extra Credit

- Refactor your project files to put common properties and/or items into a single file. ❌

## Other Notes

You did a pretty good job. After looking at your code, it made me realize that I made some stuff more complicated than it needed to be. There honestly isn't a whole lot that I can say on what you could do better, other than the comments I made above, so I do apologize for this lack luster feedback.

mmwoodfo (commented):
Good job!

PR #93, opened 2023-02-07
Assignment4 alan and grant

arose-13:
Alan Rose (@arose-13 ) and Grant Howard (@Howard404)

### DIFF
--- a/Logger.Tests/PersonTests.cs
+++ b/Logger.Tests/PersonTests.cs
@@ -1,0 +1,57 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class PersonTests
+{
+
+    [TestMethod]
+    public void CreateStudentWithMiddleName_ReturnsNonNullStudent()
+    {
+        // Assemble
+        Guid id = Guid.NewGuid();
+        Student student = new Student(id, "John", "Williams", "Towner");
+
+        // Assert
+        Assert.IsNotNull(student);
+    }
+
+    [TestMethod]
+    public void CreateStudentWithNullMiddleName_ReturnsNonNullStudent()
+    {
+        // Assemble
+        Guid id = Guid.NewGuid();
+        Student student = new Student(id, "John", "Williams", null);
+
+        // Assert
+        Assert.IsNotNull(student);
+    }
+
+    [TestMethod]
+    public void CreateStudentWithMiddleName_NameIsFullName()
+    {
+        // Assemble
+        Guid id = Guid.NewGuid();
+        Student student = new Student(id, "John", "Williams", "Towner");
+
+        // Assert
+        Assert.IsTrue(student.Name.Equals("John Towner Williams"));
+    }
+
+    [TestMethod]
+    public void CreateStudentWithNullMiddleName_NameIsFirstAndLastNames()
+    {
+        // Assemble
+        Guid id = Guid.NewGuid();
+        Student student = new Student(id, "John", "Williams", null);
+
+        // Assert
+        Assert.IsTrue(student.Name.Equals("John Williams"));
+    }
+}
--- a/Logger.Tests/PersonTests.cs
+++ b/Logger.Tests/PersonTests.cs
@@ -16,7 +16,7 @@ public class PersonTests
     {
         // Assemble
         Guid id = Guid.NewGuid();
-        Student student = new Student(id, "John", "Williams", "Towner");
+        Student student = new(id, "John", "Williams", "Towner");
 
         // Assert
         Assert.IsNotNull(student);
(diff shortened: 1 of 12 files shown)


Afterlyte:
I will review this

Logger.Tests/PersonTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        // Assert
+        Assert.IsNotNull(student);
+    }
+
+    [TestMethod]
+    public void CreateStudentWithNullMiddleName_ReturnsNonNullStudent()
+    {
+        // Assemble
+        Guid id = Guid.NewGuid();
+        Student student = new(id, "John", "Williams", null);
```
Afterlyte, line 30: An alternative to this could be to have FullName passed in as a parameter instead of the pieces of the name, that way since FullName is structured to allow MiddleName to be optional you never have to type null if you don't want a MiddleName, you just simply never provide it. However this works just as well.

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,64 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class StorageTests
```
Afterlyte, line 11: Good job testing Storage, all lines are fully covered!

Logger/BasePerson.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    // Implicit implementation so that we can create a class-specific Name implementation
+    public override string Name
+    {
+        get
+        {
+            if (string.IsNullOrWhiteSpace(FullName.MiddleName))
+                return FullName.FirstName + " " + FullName.LastName;
+            else
+                return FullName.FirstName + " " + FullName.MiddleName + " " + FullName.LastName;
```
Afterlyte, line 19: string.Format is a good way to make these return lines more readable, that way you don't have this string arithmetic going on for spaces and piecing it together.
```suggestion
                return string.Format($"{FullName.FirstName} {FullName.MiddleName} {FullName.LastName}");

```
(This is for this line as an example)

Logger/Employee.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public record class Employee : BasePerson
+{
+    public int EmployeeID { get; set; }
+    public string? Profession { get; set; }
```
Afterlyte, line 12: We probably don't want these to be publicly changeable, as those are pretty concrete pieces of information for the most part/would be dangerous to allow being messed with easily.
```suggestion
    public int EmployeeID { get; }
    public string? Profession { get; }

```

Afterlyte (changes requested):
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ❌
    - Needs comment
  - Provide a comment on the full name record on why or why not the type is immutable. ❌
    - Needs comment
    - Needs comments
  - You should consider the relationship between Student and Employee and refactor the common code shared between them. ✔
    - Good job on BasePerson for shared code
    - I like the use of FullName in book to hold the Author
  - Test that the equality behavior on these entities behaves as expected. ❌
    - Equality needs to be tested (beyond just what's done in the Storage class
  - Provide a comment on each interface method explaining why you implemented it implicitly or explicitly. ❌
    - Needs comments
- Unit test the Storage class using various entities to ensure that there are no bugs. ✔
    - (I can't really review this part haha)

## Fundamentals

- Be sure you enable:

## Extra Credit

- Refactor your project files to put common properties and/or items into a single file. ❌

Good job overall! Other than some tweaks to code and the comments/unit test stuff you nailed the requirements pretty well.

Logger/IEntity.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Logger;
 public interface IEntity
 {
-
-
+    public Guid Id { get; init; }
```
mmwoodfo, line 4:
```suggestion
    Guid Id { get; init; }
```

Logger/IEntity.cs
```diff
@@ -1,6 +1,6 @@
...
+    public string Name { get; }
```
mmwoodfo, line 5:
```suggestion
    string Name { get; }
```

Logger/BasePerson.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+
+namespace Logger;
+
+public abstract record class BasePerson : BaseEntity
+{
+    // Implicit implementation so that we can create a class-specific Name implementation
+    public override string Name
+    {
+        get
```
mmwoodfo, line 14: You could put a to string or calculated property on FullName, then you can do this formatting there!

Logger.Tests/PersonTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class PersonTests
```
mmwoodfo, line 11: What about book and student tests? You should also have tests for equality checking

PR #95, opened 2023-02-07
assignment4 1bbartels

1bbartels:


Logger/IEntity.cs
```diff
@@ -1,7 +1,6 @@
 ﻿namespace Logger;
 public interface IEntity
 {
-
-        // Place members here.
-
+    Guid Id { get; init; }
+    string? Name { get; }
```
alowe4, line 5: The question mark on string feels weird, why would there ever be an entity without a name?

alowe4 (commented):
over all look great!
______________________
Tasks
        Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.✔
        Pay special attention to the CreateLogger methods throughout. ✔

Fundamentals
    Be sure you enable:

Extra Credit

crawfordkid2:
ill be reviewing this

crawfordkid2 (commented):
Fundamentals
Be sure you enable:
Ensure there are no errors or warnings (including code analysis warnings) ❌

NOTES:
why not put all student, employee, book, and person into the same file. These classes dont have much in them, just for simplicity, you can have all the class declarations in the same file if the purpose is similar. they are all doing basically the same thing.
- add treat warnings as errors and codeanalysistreatwarnings as errors

Logger/AbstractBaseEntity.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Logger
+{
+    public abstract class AbstractBaseEntity : IEntity
```
mmwoodfo, line 9:
```suggestion
    public abstract class BaseEntity : IEntity
```
Typically you wouldn't put `abstract` in the class name

mmwoodfo (commented):
Good job overall! Looks like there were a few comments left by other students that also had some good tips.

PR #96, opened 2023-02-07
Assignment4 @mhtran23

BenagainB:


Wvvern:
Will review

Logger/FullName.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    //The FullName is a value type that could be modified through copies when called upon
+    //while keeping the main copy unchanged.
+    //The FullName record will be immutable since the changes will be performed on copies.
+    public record struct FullName(string FirstName, string LastName, string? MiddleName = null)
+    {
+        
+        public string FirstName { get; } = FirstName ?? throw new ArgumentNullException(nameof(FirstName));
+        public string LastName { get; } = LastName ?? throw new ArgumentNullException(nameof(LastName));
+        public string MiddleName { get; } = MiddleName;
```
Wvvern, line 11:
```suggestion
        public string? MiddleName { get; } = MiddleName;
```
MiddleName should be marked as nullable - since that's its default value. This also gets rid of the nullable warning.

Logger/FullName.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Logger
+{
+    //The FullName is a value type that could be modified through copies when called upon
+    //while keeping the main copy unchanged.
+    //The FullName record will be immutable since the changes will be performed on copies.
+    public record struct FullName(string FirstName, string LastName, string? MiddleName = null)
```
Wvvern, line 6:
```suggestion
    public readonly record struct FullName(string FirstName, string LastName, string? MiddleName = null)
```
Our instructor recommends always making structs readonly. VS also recommends setting it to readonly since it's all getters anyway.

Logger/Person.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Logger
+{
+    public record Person : IEntity
```
Wvvern, line 3: The requirements specify an abstract record class (a base class) that should implement IEntity.
Person should extend that abstract base class instead.
The Guid Id could also be moved into that abstract base class.

Logger/Records.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace Logger
+{   //All the records are implemented implicitly so it wouldn't conflict with values with the same class
+    public record Book : IEntity
```
Wvvern, line 3:
```suggestion
    public record class Book(string Title, FullName Author, string ISBN) : RecordsEntity 
```
Define the members of book within the record class header, and then override their default implementations to null check.
Make sure to always specify what type of record it is after record.

Logger/RecordsEntity.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger
+{
+    public abstract class RecordsEntity : IEntity
```
Wvvern, line 3: The abstract base class isn't used anywhere?
Make Book and Person extend it.

Logger/Person.cs
Wvvern, line 3:
```suggestion
    public record class Person(FullName FullName) : RecordsEntity 
```

Logger/Records.cs
```diff
@@ -0,0 +1,24 @@
...
+        public FullName Author { get; init; }
+        public string ISBN { get; init; } = string.Empty;
+
+        public string Name => $"{Title}, Author: {Author.Name}, ISBN: {ISBN}";
+
+    }
+
+    public record Student : Person
+    {
+        public int StudentID;
```
Wvvern, line 16:
```suggestion
    public record class Student(int StudentID) : Person
    {
        public int StudentID { get; } = StudentID;
```

Logger/Records.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public record Employee : Person
+    {
+        public string Employer { get; init; } = string.Empty;
```
Wvvern, line 21:
```suggestion
    public record class Employee(FullName Employer) : Person
    {
        public FullName Employer { get; init; } = Employer;
```

Logger/Person.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        public Guid Id { get; init; }
+        public FullName FullName { get; init; }
+
+        string IEntity.Name => FullName.Name;
```
Wvvern, line 8: Missing explicit/implicit explanation here.

Wvvern (commented):
### Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Provide a comment on the full name record on why or why not the type is immutable. ✔
  - You should consider the relationship between Student and Employee and refactor the common code shared between them. ✔
  - Test that the equality behavior on these entities behaves as expected.  ✔
  - Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.
- Unit test the Storage class using various entities to ensure that there are no bugs.

### Fundamentals

- Be sure you enable:
- **Ensure there are no errors or warnings (including code analysis warnings)** ✔?
- **Choose simplicity over complexity** ✔?

### Extra Credit

- Refactor your project files to put common properties and/or items into a single file. ✔

Logger/FullName.cs
BenagainB: We are straight-up lifting this recommendation and including in the PR.

Logger/Person.cs
BenagainB: Added explanation to PR.

mhtran23: will address this. thank you

Logger/Person.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Logger
+{
+
+    //implemented implicitely so other entities could use it
+    public record class Person : IEntity
+
+    {
+        public Guid Id { get; init; }
+        public FullName FullName { get; init; }
```
mmwoodfo, line 9:
```suggestion
        public FullName FullName { get; }
```
This does not need an init setter on it.

Logger/FullName.cs
```diff
@@ -0,0 +1,17 @@
...
+    public readonly record struct FullName(string FirstName, string LastName, string? MiddleName = null)
+
+    {
+        
+        public string FirstName { get; } = FirstName ?? throw new ArgumentNullException(nameof(FirstName));
+        public string LastName { get; } = LastName ?? throw new ArgumentNullException(nameof(LastName));
+        public string? MiddleName { get; } = MiddleName;
+
+
+        public string Name => $"{FirstName} {MiddleName} {LastName}";
```
mmwoodfo, line 15: You could end up with two spaces between first and last name if your middle name is null.

Logger/Person.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Logger
+{
+
+    //implemented implicitely so other entities could use it
+    public record class Person : IEntity
```
mmwoodfo, line 5:
```suggestion
    public record class Person : BaseEntity
```

Logger/Records.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Logger
+{   //All the records are implemented implicitly so it wouldn't conflict with values with the same class
+
+    public record class Book : IEntity
```
mmwoodfo, line 4: These records should all be split into their own files.

Logger/Records.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public record class Student : Person
+    {
+        public int StudentID;
+    }
+
+
+    public record class Employee : Person
+    {
+        public string Employer { get; init; } = string.Empty;
```
mmwoodfo, line 24: Rather than setting this to empty by default, set it in either a constructor or allow it to be nullable if you would expect this value to be able to hold null.

Logger/Records.cs
```diff
@@ -0,0 +1,27 @@
...
+        public string ISBN { get; init; } = string.Empty;
+
+        public string Name => $"{Title}, Author: {Author.Name}, ISBN: {ISBN}";
+
+    }
+
+
+    public record class Student : Person
+    {
+        public int StudentID;
```
mmwoodfo, line 18: Consider using the ID on entity?

Logger/Records.cs
mmwoodfo, line 4:
```suggestion
    public record class Book : BaseEntity
```

Logger.Tests/BaseEntityTest.cs
```diff
@@ -0,0 +1,75 @@
...
+                Title = bookOne.Title,
+                Author = bookOne.Author,
+                ISBN = bookOne.ISBN
+            };
+
+            Assert.AreEqual(bookOne, bookTwo);
+        }
+
+        [TestMethod]
+        public void Employee_SameProperties_AreEqual()
```
mmwoodfo, line 55: Maybe add some tests to check for when they are not equal aswell

Logger.Tests/RecordsEntityTest.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests
+{
+    [TestClass]
+    public class RecordsEntityTest
```
mmwoodfo, line 6: Split these tests up into their own files, one for student, one for book and one for employee

PR #97, opened 2023-02-07
Assignment 4

DaisyPalencia:
Pairing with @AnnikaStachowiak.


Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        Storage storage = new Storage();
+
+        storage.Add(book);
+        storage.Remove(book);
+        
+        Assert.IsFalse(storage.Contains(book));
+    }
+
+}
+
```
jdriscoll1, line 34: You should add a test for Get

Logger/Entities.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger;
+//interface implemented implicitly
+public abstract class Entities : IEntity
+{
+    public abstract string Name { get; }
+    public Guid Id { get; init; }
+}
```
jdriscoll1, line 7: This is the record that should be overriden by employee, book and student. Also, perhaps set the Globally Universal ID here?

Logger/Record.cs
```diff
@@ -0,0 +1,54 @@
...
+    //implicitly implements IEntity; more control 
+    public record Book(string? title, string? author, int isbn) : IEntity
+    {
+
+        public string title { get; } = title ?? throw new ArgumentNullException(nameof(title));
+        public string author { get; } = author ?? throw new ArgumentNullException(nameof(author));
+        public int isbn { get; } = isbn;
+
+        public Guid Id { get => throw new NotImplementedException(); init => throw new NotImplementedException(); }
+        public string Name => throw new NotImplementedException();
```
jdriscoll1, line 26: Don't throw an implemented exception for the name

Logger/Record.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public string bookName => "${title} {author} {isbn}";
```
jdriscoll1, line 28: Instead, use this functionality in the name variable

Logger/Record.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using Logger;
+using static System.Runtime.InteropServices.OptionalAttribute;
+namespace Logger;
+
+public class Record
+{
+    //We decided to do a reference type because it is better for storing more values
```
jdriscoll1, line 8: Get more specific, what about reference type makes it able to store more values and how is this better than value type

Logger/Record.cs
jdriscoll1: And then get rid of bookName, this applies with student and employee as well

Logger/Record.cs
```diff
@@ -0,0 +1,54 @@
...
+        public string bookName => "${title} {author} {isbn}";
+    }
+    //implicitly implements IEntity; more control
+    public record Student(int studentID, string? name, int grade) : IEntity
+    {
+        public int studentID { get; } = studentID;
+        public string name { get; } = name ?? throw new ArgumentNullException(nameof(name));
+        public int grade { get; } = grade;
+
+        public Guid Id { get => throw new NotImplementedException(); init => throw new NotImplementedException(); }
```
jdriscoll1, line 37: Don't throw a not implemented exception, set the Guid in the entity and then don't have it set here

jdriscoll1 (commented):
Define a full name record (first, last, middle) handling optional and null appropriately. ❌
There is not a full name record. However, there is a Person record with first last and middle. It seems unlikely that there would be a person without a first or last name. Do not make first and last name nullable, or if you do, explain why you have chosen for it to be nullable. 
Provide a comment on the full name record on why you selected to define a value or a reference type and ❌

There is no FullName record, only a person record. Also, although there is a comment about why you chose reference type, there is not enough reason given. Why is it better for storing more values? Why does Full Name need to store more values? Moreover, the reasoning given seems incorrect. One of the biggest differences between value and reference type is how the two are copied and compared, you should detail how the comparison operations are impacted and the advantages of it being a reference type through the perspective of comparing it to a different value type.  

I see a Person Record with the attributes of a FullName, but I do not see a Full Name record nor do I see an explanation as to why the person record is mutable

Define book, student, and employee records - all with entity capabilities. ❌

The records have IEntity capabilities, not Entity capabilities. 

Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ❌

I notice you have both a particular name for each record, for example employeeName and throw a “not implemented exception”. You should instead run the employeeName code for the Name variable for each record. 

Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ❌

You need to provide better reasoning for why you chose to implement it implicitly. How specifically does it give you better control and in what way is this advantageous to explicitly implementing it?

You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌

The code is not refactored so that common code is not duplicated. An example of a way to refactor the code is to perhaps create a Person class from which they both inherit. 

Given the properties on the full name record, you should consider which entities it makes sense to use it. ❌

FullName is not used inside of other records

There is not test that asserts whether two entities are equal 

Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.❌

You need to provide better reasoning for why you chose to implement it implicitly. How specifically does it give you better control and in what way is this advantageous to explicitly implimenting it?

Unit test the Storage class using various entities to ensure that there are no bugs. ❌ 
You do not test if the Get command works 

Fundamentals
Be sure you enable:
Ensure there are no errors or warnings (including code analysis warnings) ❌
Ran into Mark as CLSCompliant when I turned on including code analysis warnings, also _ between words in Test methods (take this one with a grain of salt, I don’t know the weight of CodeAnalysis bugs, although it generally is better to stay safe and get rid of them)

FullName record does not exist and thus cannot be unit tested 

Extra Credit
Refactor your project files to put common properties and/or items into a single file. ✔
All records are put together 

Overall, there are definitely improvements that can be made. Having a FullName record and Person Record which uses this record would make the code a lot cleaner. Getting rid of the not implemented exceptions on the name value would clean it up a lot. Give more specific reasoning as to why you chose to implicitly override methods.

Logger/Record.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using Logger;
+using static System.Runtime.InteropServices.OptionalAttribute;
+namespace Logger;
+
+public class Record
+{
+    //We decided to do a reference type because it is better for storing more values
+    public record Person(string? firstName, string? lastName, string? middleName)
```
eroadhouse, line 9: Curious as to why this was changed when we were given a FullName class?

Logger/Record.cs
eroadhouse, line 26: Why would you throw NotImplementedExceptions on lines 25 and 26 here?

eroadhouse (commented):
Define a full name record (first, last, middle) handling optional and null appropriately. ❌
Define book, student, and employee records - all with entity capabilities. ❌
Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ❌
Provide a comment on each interface member in each entity explaining why you implemented it implicitly or explicitly. ❌
You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌

Be sure you enable:

Logger.Tests/StorageTests.cs
DaisyPalencia: You're right, thank you! Added test for Get!

Logger/Record.cs
DaisyPalencia: Changed! Thank you

Logger/Record.cs
DaisyPalencia: Added a lengthier comment with more explanation, thanks!

Logger/Record.cs
DaisyPalencia: fixed!

Logger/Record.cs
DaisyPalencia: Updated

Logger/Record.cs
DaisyPalencia: changed

Logger/Record.cs
```diff
@@ -0,0 +1,56 @@
...
+using static System.Runtime.InteropServices.OptionalAttribute;
+namespace Logger;
+
+
+/*
+ The reason we implemented our FullName record in a class
+called records is so we could keep all of our records together.
+It makes it easier to navigate and understand our program
+from an outside perspective.  
+*/
```
mmwoodfo, line 11: Actually, I'd argue it makes it harder. I initially couldn't find your FullName record because there wasn't a file in the directory with that name, leading me to believe it didn't exist.
If you want to group things in this manor, I'd suggest splitting each record into it's own file and then grouping those files into a folder called 'records', but don't ever declare multiple classes or records inside the same file.

Logger/Record.cs
```diff
@@ -0,0 +1,56 @@
...
+public class Record
```
mmwoodfo, line 12: This class should not be here, the records should not sit inside a parent class.

Logger/Record.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        public string firstName { get; } = firstName ?? throw new ArgumentNullException(nameof(firstName));
+        public string lastName { get; } = lastName ?? throw new ArgumentNullException(nameof(lastName));
+        public string middleName { get; } = middleName ?? throw new ArgumentNullException(nameof(middleName));
+
+    }
+    //implicitly overrding Name in Entity since we are only using Name.
+    public record Person(FullName FullName) : Entity
+    {
+         public override string Name { get => FullName.ToString(); }
```
mmwoodfo, line 29:
```suggestion
         public override string Name { get => FullName.ToString(); }
```
What would you expect the `ToString` method to return here? You should consider adding a test for this.

Logger/IEntity.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger;
+
+public interface IEntity
+{
+    public Guid Id { get; init; }
+    public string Name { get; }
```
mmwoodfo, line 6:
```suggestion
    Guid Id { get; init; }
    string Name { get; }
```

Logger/Record.cs
```diff
@@ -0,0 +1,56 @@
...
+
+         
+    }
+    //implicitly overrding Name in Entity since we are only using Name.
+    public record Book(string? title, string? author, int isbn) : Entity
+    {
+
+        public override string Name { get; } = title ?? throw new ArgumentNullException(nameof(title));
```
mmwoodfo, line 37: Why throw a null exception here if you are marking the title as being nullable? Marking it as nullable is telling the person passing in the title that null is an acceptable value.

Logger/Record.cs
```diff
@@ -0,0 +1,56 @@
...
+        public string author { get; } = author ?? throw new ArgumentNullException(nameof(author));
```
mmwoodfo, line 38: Same here, you are marking author as nullable, but then throwing an exception when you get a null value.

Logger/Record.cs
```diff
@@ -0,0 +1,56 @@
...
+        public int isbn { get; } = isbn;
+    }
+    //implicitly overrding Name in Entity since we are only using Name.
+    public record Student(int studentId, FullName FullName, int grade) : Entity
+    {
+
+        public int studentId { get; } = studentId;
+        public override string Name { get => FullName.ToString(); }
```
mmwoodfo, line 46: Name for employee and student could probably be moved to a shared record, such as `Person` since they are doing the exact same thing.

Logger.Tests/RecordTests.cs
```diff
@@ -0,0 +1,94 @@
...
+    [TestMethod]
+    [ExpectedException(typeof(ArgumentNullException))]
+    public void CreateEmployeeNotSuccess()
+    {
+        Record.Employee jim = new Record.Employee(null!, null!);
+
+        Assert.IsNull(jim);
+
+    }
+    [TestMethod]
```
mmwoodfo, line 75: You should have equality tests for all your records, as well as inequality tests

PR #100, opened 2023-02-07
Assignment 4 - Hunter Thompson & Hunter Allen

StrangeRanger:
@hallen10


GeorgeKoralski007 (changes requested):
I will review

JessicaH46:

I think your code looks really good, your tests are very thorough and well written. Also we were having problems on our code with the inheritance but I saw that you used sealed, which I think works great and was helpful even though I was reviewing your code! Great job!

GeorgeKoralski007:
**Tasks**
**Fundamentals**
Be sure you enable:
**Extra Credit**
Returns the logger classes type that is to be created. ❌✔
================================================================
Your project Builds
2 of your Tests do not pass (AreSame_Book and Create_Student_WithMiddleName)

Logger/Entity.cs
```diff
@@ -0,0 +1,18 @@
...
+ * The interface is implemented implicitly because it doesn't make sense otherwise.
+ * Additionally, the compiler would throw an error if I tried to implement it implicitly
+ */
+public abstract record class Entity : IEntity
+{
+    public Guid Id { get; init; }
+    public abstract string Name { get; set; }
+
+    // Constructor.
+    protected Entity(Guid id, string? name = null)
```
GeorgeKoralski007, line 13: param name is not used

Logger.Tests/BookTest.cs
```diff
@@ -0,0 +1,25 @@
...
+        Assert.AreEqual("J.R.R. Tolkien", book.Author);
+        Assert.AreEqual("The Hobbit by J.R.R. Tolkien", book.ToString());
+    }
+    
+    [TestMethod]
+    public void AreSame_Book()
+    {
+        var book1 = new Book("The Hobbit", "J.R.R. Tolkien");
+        var book2 = new Book("The Hobbit", "J.R.R. Tolkien");
+        Assert.IsTrue(book1.Equals(book2));
```
GeorgeKoralski007, line 23: test is failing because id for each record is different. You can compare Assert.AreEqual(book1.ToString(),book2.ToString());

Logger.Tests/StudentTest.cs
```diff
@@ -0,0 +1,38 @@
...
+        Assert.IsNull(student.MiddleName);
+    }
+    
+    [TestMethod]
+    public void Create_Student_WithMiddleName()
+    {
+        Student student = new Student(Guid.NewGuid(), "John Michael Doe");
+        Assert.AreEqual("John Doe Michael", student.Name);
+        Assert.AreEqual("John", student.FirstName);
+        Assert.AreEqual("Doe", student.LastName);
```
GeorgeKoralski007, line 25: Your test is failing because your implementation of the Name setter incorrectly assigned middle and last name.

Logger/Student.cs
```diff
@@ -0,0 +1,57 @@
...
+     * wouldn't make sense to implement the the IEntity interface when we are extending Entity.
+     */
+    public sealed override string Name
+    {
+        get { return EntityFullName.ToString(); }
+        set
+        {
+            string[] array = value.Split(' ');
+            string firstName = array[0];
+            string lastName = array[1];
```
GeorgeKoralski007, line 45: We should assume Name is not always (valid with 2 or 3 names). Otherwise accessing array may be out of range.
Validate array.Count before that

Logger/Student.cs
```diff
@@ -0,0 +1,57 @@
...
+            string middleName = array.Length > 2 ? array[2] : null; // TODO: MAKE THIS BETTER
```
GeorgeKoralski007, line 46: Correct input value should be:
"First Last" or "First Middle Last"
In your case you assign middleName to Last, hence your unit test fails.
My suggestion is:
if (array.Length == 3)
{
  middleName = array[1];
  lastName = array[2];
}

Logger/Entity.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+namespace Logger;
+
+/**
+ * The interface is implemented implicitly because it doesn't make sense otherwise.
+ * Additionally, the compiler would throw an error if I tried to implement it implicitly
+ */
+public abstract record class Entity : IEntity
+{
+    public Guid Id { get; init; }
```
GeorgeKoralski007, line 10: You can initialize the Id to be Guid.NewGuid() instead of having each subclass do the same work.

Logger/Employee.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Logger;
+
+/**
+ * TODO: We weren't sure how to refactor the similarities of the Employee and Student class,
+ * because as it stands, if we did that, we could just add everything into pretty much a
+ * single file, which doesn't really make sense if, if both classes represent separate objects.
+ * Please give use some input on what you think we should do
+ */
+
+public record class Employee : Entity
```
GeorgeKoralski007, line 11: Employee and Student are the same and you have code duplication.
You can have specific property that applies only for one of the classes.
All common properties can be in a base Person class for example.

GeorgeKoralski007 (changes requested):
Please review my comments

Logger/Entity.cs
StrangeRanger: Looking back at it, I could have changed it were the name parameter was never specified, then convert the constructor into a primary constructor. 
Thanks for pointing that out for me.

Logger.Tests/BookTest.cs
StrangeRanger: No longer applicable

Logger.Tests/StudentTest.cs
StrangeRanger: No longer applicable

Logger/Student.cs
StrangeRanger: No longer applicable

Logger/Entity.cs
StrangeRanger: No longer applicable

Logger/Employee.cs
StrangeRanger: Thank you for your input here, and answering the question I left.

Logger.Tests/BookTest.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        FullName fullName = new FullName("J.R.R.", "Tolkien");
+        Book book = new Book(Guid.NewGuid(), fullName, "The Hobbit");
+        
+        Assert.AreEqual("The Hobbit", book.Title);
+        Assert.AreEqual("J.R.R. Tolkien", book.Name);
+    }
+    
+    [TestMethod]
+    public void AreSame_Book()
```
mmwoodfo, line 19: Maybe also check for inequality!

Logger/Entity.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+namespace Logger;
+
+/**
+ * The interface is implemented implicitly because it doesn't make sense otherwise.
+ * Additionally, the compiler would throw an error if I tried to implement it implicitly
+ */
+public abstract record class Entity(Guid Id) : IEntity
```
mmwoodfo, line 8:
```suggestion
public abstract record class BaseEntity(Guid Id) : IEntity
```

Logger/IEntity.cs
```diff
@@ -1,7 +1,8 @@
...
+using System;
+namespace Logger;
+
 public interface IEntity
 {
-
-        // Place members here.
-
+    public Guid Id { get; init; }
+    public string Name { get; }
```
mmwoodfo, line 7:
```suggestion
    Guid Id { get; init; }
    string Name { get; }
```

Logger/Person.cs
```diff
@@ -0,0 +1,39 @@
...
+public record class Person : Entity
+{
+    public Person(Guid id, FullName fullName) : base(id)
+    {
+        Id = id;
+        EntityFullName = fullName;
+    }
+    
+    private FullName EntityFullName { get; }
+    public string FirstName
```
mmwoodfo, line 13: Do you think it would make more sense to just make the FullName property public rather than having these extra methods?

StrangeRanger: Now that I think about it, probably

Logger/IEntity.cs
StrangeRanger: I get the ID, but why would we want the make Name private?

Logger/Entity.cs
StrangeRanger: If it's a base element, is it ideal to always prefix it as such?

PR #101, opened 2023-02-07
Assignment 4 with Vera Korchemnaya

AndrewKorchemniy:


Logger/Book.cs
```diff
@@ -0,0 +1,22 @@
...
+public record class Book(string Title, string Author, string ISBN) : BaseEntity, IEquatable<Book>
+{
+    public string Title { get; } = Title ?? throw new ArgumentNullException(nameof(Title));
+    public string Author { get; } = Author ?? throw new ArgumentNullException(nameof(Author));
+    public string ISBN { get; } = ISBN ?? throw new ArgumentNullException(nameof(ISBN));
+
+    // Implemented implicitely because we don't want to hide this member from the users of the class
+    public override string Name { get => $"{Title} by {Author} ({ISBN})"; }
+
+    public virtual bool Equals(Book? other)
```
FenrisStargazer, line 12: I thought records provided equals methods? And why compare only off the ISBN? Theoretically, yes, it would mean there would never be overlap because all books have a distinct ISBN, but I feel like there is an argument to be made for perhaps an older book that may not have an ISBN or something.

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Logger;
+
+public abstract record class BaseEntity : IEntity
```
FenrisStargazer, line 3: Besides for the sake of burying the Id property and not having to implement it explicitly again and again, what purpose exactly does this method serve? What makes it necessary in the sense the Person class couldn't just implement IEntity directly without this class?

Logger.Tests/Logger.Tests.csproj
```diff
@@ -18,6 +18,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Logger\Logger.csproj" />
+    <Using Include ="Microsoft.VisualStudio.TestTools.UnitTesting"/>
```
FenrisStargazer, line 21: I'm presuming this is so it doesn't have to add that using statement to all the test files? Very clever!

FenrisStargazer (commented):
**Tasks**
_I would say technically yes, because of the comment in the Person class and in Book, but there's nothing in Employee or Student because they don't directly have Name or Id from the interface._
_Again, technically yes? Because they're both basically empty except for a fullName which is it's own record class._
Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.)❌?✔
_No comments, but maybe that just means there was nothing wrong, I can't really judge._

**Fundamentals**
Be sure you enable:

**Extra Credit**
Refactor your project files to put common properties and/or items into a single file. ❌?✔
_Given the state of Employee and Student, arguably maybe a little bit? But I'm not really sure._

All tests pass.
Overall I think you did really, really well and made a lot of good points or took interesting approaches that I hadn't considered. I'm not sure my feedback will be the most helpful to you because it didn't seem like there was much to critique, but I definitely feel like I learned something from the way you implemented some of this stuff and the reasonings you gave so good job!

mhtran23:
Choose a thoughtful approach for the value of Name for each type of entity. In all cases, there should be no backing field for Name (not even one generated by an automatically implemented property) - consider a calculated property. ❌
You should consider the relationship between Student and Employee and refactor the common code shared between them. ❌
Fundamentals
Be sure you enable:

The only real issue I see is that there isn't any real way to differentiate between Employee and Student. Give them extra treats such as or something similar:

public record Student : Person
    {
        public int StudentID;
    }

    public record Employee : Person
    {
        public string Employer { get; init; } = string.Empty;
    }

Logger/Book.cs
Wvvern: Yes, records do provide equals methods. However, we chose to override the equals method because the Book class derives from BaseEntity, giving it a unique Guid ID. So, even if a book was identical on all properties, it would still return false.
Our Book implementation doesn't support books without ISBN, but we think that's a reasonable requirement.

Logger/BaseEntity.cs
Wvvern: We think burying the Guid Id property is a significant enough reason. If Person was also a base entity class, it would seem repe |
Furthermore, we didn't want two abstract base classes. Having one for a book which doesn't fit into the person category.

Wvvern: We think buring the Guid Id property is a significant enough reason. Otherwise, we're duplicating the implementation of the Guid Id. Having everything inherit a single global abstract base class reduces complexity.

Logger/Student.cs
```diff
@@ -0,0 +1,5 @@
+﻿using System;
+
+namespace Logger;
+
+public record class Student(FullName FullName) : Person(FullName);
```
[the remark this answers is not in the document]
Wvvern: Yea, this is something we decided to expand on today. Thanks

mmwoodfo (commented):
Great job, my only comment would be to just make sure to treat warnings as errors is on as well as a few other analyzers. These are not always set by default when you pull the project

PR #102, opened 2023-02-07
Assignment 4 With @GeorgeKoralski007

crawfordkid2:


slanante:
Will review.

BenagainB:
Fundamentals
Be sure you enable:
Extra Credit
Refactor your project files to put common properties and/or items into a single file. ✔
Convert ILogger.CreateLogger to a generic method. ✔
Returns the logger classes type that is to be created. ✔
Has a type parameter for the configuration type ✔

Man, I really think you nailed this one! Only changes I'd suggest would be to go back through your files and see what comments can just be ripped out (FullNameRecord.cs is absolutely packed with comments). Other than that, it looks excellent and I may have to go back now and make sure my generic method is written as cleanly as yours.

Logger/FullNameRecord.cs
```diff
@@ -0,0 +1,42 @@
...
+    //this is the record that turns someones full name into a name, (First last middle) = name using our abstract class
+    public record Someone(FullName Full): AbstractEntity
+    {
+        public override string Name { get => Full.ToString(); }
+    }
+    //the employee record uses fullname and the someone record to assign a full name and make them someone implicitly
+    public record Employee(FullName Full, string Position, string Company) : Someone(Full);
+    //the student record is very similar to the employee but is slightly different with what fields are needed.
+    public record Student(FullName Full, string Degree, string School) : Someone(Full);
+}
```
slanante, line 42: I would recommend placing these classes in their own files. The FullNameRecord file does not have a name that would suggest the Emplyoee, Student, and Book classes are contained within. I would also suggest placing the Someone class inside its own class file as well.

slanante: You are also not setting attributes for the Book, Employee or Student class. You are passing parameters in Book for example, but the only parameter that is saved by a book object is Name. You have not set Author or ISBN as attributes of Book and these parameters will not be saved into an instance of Book. Equality checks will fail for your current implementation of these classes but with a correct implementation they should pass. Below is a screenshot of a test I added on to your code. 
image

slanante (commented):
Overall, good job. There is a need to edit the Book, Employee and Student records as their checks for equality will fail with the current implementation. Separating these same classes into their own files will also help improve the organization and help ease future work (such as future edits) on the code for other people.

Define book, student, and employee records - all with entity capabilities. ❌ (This needs revision)
Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.✔
Fundamentals
Be sure you enable:

Logger/FullNameRecord.cs
```diff
@@ -0,0 +1,42 @@
...
+                 // Instead if we made these mutable objects, there is a chance we would overwrite something and "delete" a previous record.
+
+{
+
+    //here is the FullName struct, makes sure each full name has a first and last, middle can be null because not all 
+    //people may have a middle name
+    //struct allows for more freedom when using this for all of our other records\
+    // this is an implicit way of letting the compiler know that each full name always has a first and last name while allowing middle name to be null
+
+    public readonly record struct FullName(string First, string Last, string? Middle = null)
```
mmwoodfo, line 11: These records should be in their own classes

Logger/FullNameRecord.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        public override string Name { get => Title; }
+
+    }
+
+
+    //this is the record that turns someones full name into a name, (First last middle) = name using our abstract class
+    public record Someone(FullName Full): AbstractEntity
+    {
+        public override string Name { get => Full.ToString(); }
```
mmwoodfo, line 36: What do you expect ToString to print here? Consider adding a test for this.

Logger/AbstractEntity.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger
+{
+    public abstract record class AbstractEntity : IEntity
```
mmwoodfo, line 3:
```suggestion
    public abstract record class BaseEntity : IEntity
```

Logger/IEntity.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger;
+public interface IEntity
+{
+
+    // Place members here.
+    public Guid Id { get; init; }
+    public string Name { get; }
```
mmwoodfo, line 7:
```suggestion
    Guid Id { get; init; }
    string Name { get; }
```

Logger.Tests/NameTests.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests
```
mmwoodfo, line 3: It looks like you are missing equality/inequality tests for your records

PR #107, opened 2023-02-07
Assignment4 -Victor Korn

vicdotexe:


randypipkins (commented):
All code compiles and runs successfully and all unit tests succeed. Only suggestion I might make would be to make sure your { } are on separate lines when declaring get; and set; to improve code readability

Logger.Tests/TestEntity.cs
```diff
@@ -0,0 +1,70 @@
...
+        var last = "Korn";
+        var middle = "Ernest";
+        var fullName = new FullName(first, last, middle);
+        var person = new Person(fullName);
+        Assert.AreEqual(person.Name, $"{first} {middle} {last}");
+    }
+
+    [TestMethod]
+    public void StudentTest()
+    {
```
mmoran314, line 64: Maybe add tests for the added parameters for each record, like major in student.

Logger/Employee.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Logger;
+
+public record class Employee(FullName FullName, string Role) : Person(FullName)
+{
```
mmoran314, line 4: I think that these classes need properties for the added parameters i.e. for Role in the Employee class and Major in student class

mmoran314 (commented):
Fundamentals
Be sure you enable:

I think it looks good. I think that you might need to add to some of the records for the new parameters that they have so that everything gets implemented when creating the new instance of the object, adding tests for these as well.

mmwoodfo:
> All code compiles and runs successfully and all unit tests succeed. Only suggestion I might make would be to make sure your { } are on separate lines when declaring get; and set; to improve code readability

If the getter and setters are just being declared, they should stay inline. Is there a specific area in the code you were referring to here?

Logger/Employee.cs
vicdotexe: Like I need to override the ones that a record automatically generates? I don't understand why.

Logger/FullName.cs
```diff
@@ -0,0 +1,20 @@
...
+//and safeguard against assuming the changes have affected any copies.
+public readonly record struct FullName(string FirstName, string LastName, string? MiddleName = null)
+{
+    public string FirstName { get; } = FirstName ?? throw new ArgumentNullException(nameof(FirstName));
+    public string LastName { get; } = LastName ?? throw new ArgumentNullException(nameof(LastName));
+    public string? MiddleName { get; } = MiddleName;
+
+    public override string ToString()
+    {
+        return MiddleName == null ? $"{FirstName} {LastName}" : $"{FirstName} {MiddleName} {LastName}";
```
mmwoodfo, line 18: Nice, you can also use something called pattern matching here for a more natural feel when reading this statement (note, this pattern matching only works for equality when checking enums and null values)
```suggestion
        return MiddleName is null ? $"{FirstName} {LastName}" : $"{FirstName} {MiddleName} {LastName}";
```

Logger/IEntity.cs
```diff
@@ -1,7 +1,6 @@
 ﻿namespace Logger;
 public interface IEntity
 {
-
-        // Place members here.
-
+    Guid Id { get; init; }
+    string? Name { get; }
```
mmwoodfo, line 5: I don't think you need to make name nullable here.

Logger/Employee.cs
mmwoodfo: You do not need to declare parameters unless you intend on implementing any kind of getter/setter logic on the properties.

Logger.Tests/TestEntity.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Logger.Tests;
+
+[TestClass]
+public class TestEntity
+{
+    [TestMethod]
+    public void Book_CompareDifferentIds_NotEqual()
```
mmwoodfo, line 9: Good job adding inequality checks as well!

PR #109, opened 2023-02-12
Assignment5

FenrisStargazer:

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTest.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTest.cs
@@ -1,0 +1,52 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace GenericsHomework.Tests;
+
+[TestClass]
+public class NodeTest
+{
+    Node<int> node = new(5, null!);
+    [TestMethod]
+    public void InitializeTest()
+    {
+        Assert.AreEqual<Node<int>>(node, node.Next);
+        Assert.AreEqual<int>(5, node.Item);
+    }
+
+    [TestMethod]
+    public void ToStringTest()
+    {
+        Assert.AreEqual<string>("5", node.ToString());
+    }
+
+    [TestMethod]
+    [ExpectedException(typeof(ArgumentException))]
+    public void AppendTest_Error()
+    {
+        node.Append(5);
+    }
+
+    [TestMethod]
+    public void AppendTest_Success()
+    {
+        node.Append(3);
+        Assert.AreEqual<Node<int>>(node, node.Next.Next);
+        Assert.AreEqual<int>(3, node.Next.Item);
+    }
+
+    [TestMethod]
+    public void ExistsTest()
+    {
+        node.Append(7);
+        Assert.IsTrue(node.Exists(5));
+        Assert.IsTrue(node.Exists(7));
+        Assert.IsFalse(node.Exists(9));
+    }
+
+    [TestMethod]
+    public void ClearTest()
+    {
+        node.Clear();
+        Assert.AreEqual<Node<int>>(node, node.Next);
+        Assert.AreEqual<int>(5, node.Item);
+    }
+}
--- a/GenericsHomework/GenericsHomework/Node.cs
+++ b/GenericsHomework/GenericsHomework/Node.cs
@@ -1,0 +1,59 @@
+﻿namespace GenericsHomework;
+public class Node<T>
+{
+    private int Size { get; set; }
+    public T? Item { get; set; }
+    public Node<T> Next
+    {get; private set; }
+
+    public Node(T? item, Node<T> next)
+    {
+        Item = item;
+        Next = next ?? this;
+        Size = 1;
+    }
+
+    public override string ToString()
+    {
+        return Item!.ToString() ?? "null";
+    }
+
+    public void Append(T item)
+    {
+        if (Next.Exists(item)) throw new ArgumentException(nameof(item) + " already exists");
+        Next = new Node<T>(item, Next);
+        Size++;
+    }
+
+    public bool Exists(T item)
+    {
+        bool result = false;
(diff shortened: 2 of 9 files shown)


Nolan-Posey (commented):
Define the Node class

Fundamentals

Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔
Implement System.Collections.Generic.ICollection<T> on the Node class ❌


You absolutely nailed this assignment! I only see one issue and it is that you did not add a "Directory.Build.targets" file for all the shared project items but that is a very simple addition.

Overall, Good Job!

-Nolan Posey

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+    public Node(T? item, Node<T> next)
+    {
+        Item = item;
+        Next = next ?? this;
+        Size = 1;
+    }
+
+    public override string ToString()
+    {
+        return Item!.ToString() ?? "null";
```
mmwoodfo, line 18:
```suggestion
        return Item?.ToString() ?? "null";
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    private int Size { get; set; }
+    public T? Item { get; set; }
+    public Node<T> Next
+    {get; private set; }
+
+    public Node(T? item, Node<T> next)
+    {
+        Item = item;
+        Next = next ?? this;
```
mmwoodfo, line 12:
```suggestion
        Next = this;
```
In the constructor, next should just be set to `this`

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace GenericsHomework;
+public class Node<T>
+{
+    private int Size { get; set; }
+    public T? Item { get; set; }
+    public Node<T> Next
+    {get; private set; }
+
+    public Node(T? item, Node<T> next)
```
mmwoodfo, line 9:
```suggestion
    public Node(T item)
```
consider not allowing null items

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace GenericsHomework;
+public class Node<T>
+{
+    private int Size { get; set; }
```
mmwoodfo, line 4: Tracking this via an int is a bit weird, it might be better to track this via a method that traverses the linked list and counts it. It's weird because having it here means that each node needs to keep track of the list's size as a whole.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public override string ToString()
+    {
+        return Item!.ToString() ?? "null";
+    }
+
+    public void Append(T item)
+    {
+        if (Next.Exists(item)) throw new ArgumentException(nameof(item) + " already exists");
```
mmwoodfo, line 23:
```suggestion
        if (Exists(item)) throw new ArgumentException(nameof(item) + " already exists");
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+    {get; private set; }
+
+    public Node(T? item, Node<T> next)
+    {
+        Item = item;
+        Next = next ?? this;
+        Size = 1;
+    }
+
+    public override string ToString()
```
mmwoodfo, line 16: You can return a nullable string here, which might simplify the below code
```suggestion
    public override string? ToString()
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace GenericsHomework;
+public class Node<T>
+{
+    private int Size { get; set; }
+    public T? Item { get; set; }
+    public Node<T> Next
+    {get; private set; }
```
mmwoodfo, line 7: This should not be multilined

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+                if (current.Item is not null && current.Item.Equals(item)) result = true;
+                else current = current.Next;
+            }
+        }
+        return result;
+    }
+
+    public void Clear()
+    {
+        Next = this;
```
mmwoodfo, line 49: While yes, this will be cleaned up... You _should_ loop through all the nodes and disconnect them from each other. If there are any references held outside the list they won't be guaranteed to be cleaned up.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        bool result = false;
+        if (Item is not null && Item.Equals(item))
+        {
+            result = true;
+        }
+        else
+        {
+            Node<T> current = Next;
+            for(int i = 0; i < Size; i++)
```
mmwoodfo, line 38: Rather than using `size` here, consider using a do while loop that loops until the next value is set back to it's self.

GenericsHomework/GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,52 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace GenericsHomework.Tests;
+
+[TestClass]
+public class NodeTest
```
mmwoodfo, line 5: You should have some more tests that tests your next property and append property... E.g. something like this:
```
        [TestMethod]
        public void Insert1()
        {
            Node<int> node1 = new(1);
            var node2 = node1.Append(2);
            Assert.AreEqual(node2, node1.Next);
            Assert.AreEqual(node1, node2.Next);
        }
```

PR #110, opened 2023-02-12
Assignment5-VictorKorn

vicdotexe:


j0shh3ss (commented):
Define the Node class✔

Tips-
Nicely done! It seems like you hit all the checkmarks. I haven't made my Next property nullable so I have no advice on doing it, but remember to do so!
Also remember to add your POM file from the Directory.Build.props file as well.

crawfordkid2 (commented):
Define the Node class

Fundamentals

Notes: 
double check to add generics to your Assert.AreEqual
make sure next is non nullable,

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,141 @@
...
+        public Node(T value)
+        {
+            Value = value;
+            Next = this;
+        }
+
+        public Node<T> Append(T value)
+        {
+            if (Exists(value))
+                throw new ArgumentException("Cannot add same value more than once", nameof(value));
```
mmwoodfo, line 21:
```suggestion
            if (Exists(value)) throw new ArgumentException("Cannot add same value more than once", nameof(value));
```
This is a minor nit pick, but I'd suggest either keeping this inline or using curly brackets once you start to multi-line it.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            var next = Next;
+            Next = new Node<T>(value)
+            {
+                Next = next
+            };
+            return Next;
+
```
mmwoodfo, line 29: while this is correct, it may be easier to read if you break it up a bit more with more specific variable names
```suggestion
       Node<T> newNode = new(value);
       newNode.Next = Next;
       Next = newNode;
       return newNode;
```

Just easier to track what the new next node is vs the currently set next node.

PR #111, opened 2023-02-13
Assignment5 @avanchu253

randypipkins:


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        Next = this;
+    }
+
+    public TVal Val
+    {
+        get;
+        set;
+    }
+
+    public Node<TVal> Next
```
DaisyPalencia, line 17: I recommend checking to see if the next property is non-nullable.

DaisyPalencia (commented):
Instructions
Define the Node class
Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ❌

Fundamentals

Good job! If you have any questions or if you have something and I marked it as wrong please let me know!

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,96 @@
+using GenericsHomework;
```
DaisyPalencia, line 1: Good Job with your tests! I think I will go back and add more tests to my own after looking over yours!

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace GenericsHomework;
+
+public class Node<TVal> where TVal : notnull
+{
+    public Node(TVal val)
+    {
+        Val = val ?? throw new ArgumentNullException($"Node<{typeof(TVal)}> value cannot be null");
```
mmwoodfo, line 7: Since your value has a public setter, if you're going to do validation, it should be done on your property nnot here.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace GenericsHomework;
+
+public class Node<TVal> where TVal : notnull
```
mmwoodfo, line 3:
```suggestion
public class Node<T>
```
You don't need this here, also `T` is the commonly used identifier for generics.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    public Node(TVal val)
```
mmwoodfo, line 5:
```suggestion
    public Node(TVal value)
```
Consider using non-abbreivated names

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public void Append(TVal node)
+    {
+        if(Exists(node))
+        {
+            throw new ArgumentException("This value already exists. Value must be unique.");
+        }
+
+        Node<TVal> node1 = new(node);
```
mmwoodfo, line 30:
```suggestion
        Node<TVal> newNode = new(node);
```
maybe a better name, adding the number 1 after node is not very clear

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        if(Next == this)
```
mmwoodfo, line 32:
```
newNode.Next = Next;
Next = newNode;
```
You can do this if/else in a more straightforward way.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        set;
+    }
+
+    public Node<TVal> Next
+    {
+        get;
+        private set;
+    }
+
+    public void Append(TVal node)
```
mmwoodfo, line 23: It would be handy if the new node was returned, but it's not required.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+    /*
+     * Need only to get rid of the link from the first
+     * node because only the first node can access the subsequent nodes,
+     * thus leaving no more references to the objects created.
+     * This is proven in GenericsHomeworkTests.Node_GarbageCollection()
+     */
+
+    public void Clear()
+    {
+        Next = this;
```
mmwoodfo, line 53: While yes, this will be cleaned up... You should loop through all the nodes and disconnect them from each other. If there are any references held outside the list they won't be guaranteed to be cleaned up.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public Boolean Exists(TVal val)
+    {
+        Node<TVal> temp = this;
```
mmwoodfo, line 58:
```suggestion
        Node<TVal> node = this;
```
temp is a bad name here, as it isn't a temp variable. It's being used and updated as part of the looping logic.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        while (!temp.Val.Equals(val))
```
mmwoodfo, line 59: You should use `EqualityComparer` here

PR #112, opened 2023-02-13
Assignment5 with alowe4

jdriscoll1:
Extra Credit Attempted


GeorgeKoralski007:
I will review

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+
+            TestNode = new("myString");
+
+        }
+
+
+        [TestMethod]
+        public void CheckIfNextThrowsExceptionIfNull()
+        {
+            Assert.IsNotNull(TestNode);
```
GeorgeKoralski007, line 27: This is not testing your main code. Its testing your test (that TestNode is initialized).
You dont need to test that test framework will call TestInitialize method.
You can remove this test

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+
+
+
+        [TestMethod]
+        public void AppendNodeTest()
+        {
+            Assert.IsNotNull(TestNode);
+            TestNode.Append("Test Node");
+
+            Assert.IsTrue(TestNode.Exists("Test Node"));
```
GeorgeKoralski007, line 39: You should also test the false case when value is missing.

GeorgeKoralski007, line 39: You only test the value.
Also you should assert that both nodes are connected with .Next

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,212 @@
...
+                    return true;
+                }
+                curr = curr.Next; 
+
+            } while (curr != head);
+            return false; 
+        }
+
+
+        public override string ToString()
```
GeorgeKoralski007, line 91: Project requirement is:
Add a ToString() override that writes out the value's ToString() result. 
This means you only need to return Value.ToString() and should not walk the whole list

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,212 @@
...
+            {
+                yield return curr.Value!;
+                curr = curr.Next;
+
+            } while (curr != this);
+
+        }
+
+
+        IEnumerator<TNodeType> IEnumerable<TNodeType>.GetEnumerator()
```
GeorgeKoralski007, line 200: This method and the one above has duplicate code.
Can you call from one method the other one to avoid duplication?

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+            TestNode.Append("3");
+            TestNode.Append("4");
+            TestNode.Append("5");
+            TestNode.Append("6");
+            TestNode.Append("7");  
+            Assert.AreEqual<string>( "myString 7 6 5 4 3 2 1 ", TestNode.ToString()); 
+
+            
+            // Act
+            TestNode.Clear();
```
GeorgeKoralski007, line 59: Please test that TestNode.Next is the same as TestNode

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+
+
+        }
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentException))]
+        public void AssertDuplicatesCannotBeAddedNode() {
+            Assert.IsNotNull(TestNode);
+            TestNode.Append("1");
+            TestNode.Append("1");
+
```
GeorgeKoralski007, line 73: Remove all empty lines where not needed

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+        [TestMethod]
+        public void ExistsNodeTestBody()
+        {
+            // Arrange
+            Assert.IsNotNull(TestNode);
+
+            TestNode.Append("1");
+            TestNode.Append("2");
+            TestNode.Append("3");
+            Assert.IsTrue(TestNode.Exists("2"));
```
GeorgeKoralski007, line 97: Also test when Exists will return false.

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+        [TestMethod]
+        public void ContainsNodeTestBody()
+        {
+            // Arrange
+            Assert.IsNotNull(TestNode);
+
+            TestNode.Add("1");
+            TestNode.Add("2");
+            TestNode.Add("3");
+            Assert.IsTrue(TestNode.Contains("2"));
```
GeorgeKoralski007, line 110: Test when it will be false

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+        [TestMethod]
+        public void ConvertLinkedListToArray_StartAt0()
+        {
+            Assert.IsNotNull(TestNode);
+            TestNode.Add("1");
+            TestNode.Add("2");
+            string[] arr = {"3", "4", "5"};
+            TestNode.CopyTo(arr, 0);
+            
+            Assert.AreEqual<string>(string.Join(" ", arr), TestNode.ToString().Trim());
```
GeorgeKoralski007, line 159: You can use CollectionAssert.AreEqual instead of relying on ToString

GenericsHomework/GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,272 @@
...
+        public void ConvertLinkedListToArray_StartAt1()
+        {
+            Assert.IsNotNull(TestNode);
+            TestNode.Add("1");
+            TestNode.Add("2");
+            string[] arr = { "3", "4", "5" };
+            TestNode.CopyTo(arr, 1);
+
+            Assert.AreEqual<string>("3 myString 2", string.Join(" ", arr));
+            TestNode.Clear();
```
GeorgeKoralski007, line 174: You dont need to call Clear at the end of each test.
It will be initialized for each test anyway

GeorgeKoralski007 (changes requested):
Please review

GeorgeKoralski007:
Define the Node class
Add a ToString() override that writes out the value's ToString() result. ❌your code returns the whole list (should be a single value)
The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ❌Doesn't need to be nullable
Fundamentals

GenericsHomework/GenericsHomework/Node.cs
jdriscoll1: Thank you for the feedback. I looked for methods to rewrite the code without duplication; and, of what was attempted the current solution is the cleanest.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,190 @@
+﻿using System.Collections;
+using System.Drawing;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace GenericsHomework
+{
+    public class Node<TNodeType> : ICollection<TNodeType>
+    {
+        private Node<TNodeType>? _Next { get; set; }
```
mmwoodfo, line 9:
```suggestion
        private Node<TNodeType>? _Next;
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,190 @@
...
+            Value = value; 
+            Next = next;
+        }
+
+        public void Append(TNodeType data) {
+            if (Exists(data)) {
+                throw new ArgumentException($"{data} is already added in list");
+            }
+            Node<TNodeType> newNode = new(data, Next);
+            Console.WriteLine(newNode.Value);
```
mmwoodfo, line 59:
```suggestion
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+        }
+        bool ICollection<TNodeType>.IsReadOnly { get; } = false; 
+        public Node(TNodeType value)
+        {
+            Value = value;
+            Next = this; 
+        }
+
+        private Node(TNodeType value, Node<TNodeType> next) {
```
mmwoodfo, line 49:
```suggestion
        private Node(TNodeType value) {
```
Just set next to `this` when creating a new node.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,190 @@
...
+                array[i] = element;
+                i++; 
+            }
+            return array; 
+        }
+
+        public override string ToString()
+        {
+            return Value?.ToString() ?? "null";
+        }
```
mmwoodfo, line 93:
```suggestion
        public override string? ToString()
        {
            return Value?.ToString();
        }
```

mmwoodfo (commented):
Great job!

PR #113, opened 2023-02-13
Assignment5

Bhopss152:
JoshuaHess
j0shh3ss

### DIFF
--- a/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework.Tests/NodeTests.cs
@@ -17,8 +17,8 @@ namespace GenericsHomework
             node.Append(20);
             node.Append(30);
 
-            Assert.AreEqual(20, node.Next.Data);
-            Assert.AreEqual(30, node.Next.Next.Data);
+            Assert.AreEqual<int>(20, node.Next.Data);
+            Assert.AreEqual<int>(30, node.Next.Next.Data);
         }
 
         [TestMethod]
@@ -44,11 +44,25 @@ namespace GenericsHomework
         }
 
         [TestMethod]
+        public void TestAppendThrowsArgumentExceptionIfDataAlreadyExists()
+        {
+            // Arrange
+            var list = new Node<int>(20);
+            list.Append(30);
+
+            // Act and Assert
+            Assert.ThrowsException<ArgumentException>(() =>
+            {
+                list.Append(30);
+            });
+        }
+
+        [TestMethod]
         public void TestToStringMethod()
         {
             Node<int> node = new Node<int>(10);
 
-            Assert.AreEqual("10", node.ToString());
+            Assert.AreEqual<string>("10", node.ToString());
         }
     }
 }
--- a/GenericsHomework/GenericsHomework/Node.cs
+++ b/GenericsHomework/GenericsHomework/Node.cs
@@ -5,9 +5,12 @@
         public T? Data { get; set; }
         private int Size { get; set; }
         public Node<T> Next { get; set; }
+
         public Node(T? data)
         {
-            Data = data;
+
+            this.Data = data;
+            this.Next = this;
         }
         public override string ToString()
         {
@@ -15,7 +18,12 @@
         }
         public void Append(T data)
         {
-            Next = new Node<T>(data);
+            Node<T> cur = this;
+            while (cur.Next != this) 
+            {
+                cur = cur.Next;
+            }
+            cur.Next = new Node<T>(data); 
         }
         public void Clear()
         {
@@ -34,7 +42,8 @@
                 throw new ArgumentNullException(nameof(data));
             }
             Node<T> cur = Next;
-            while (cur != null)
+            Node<T> head = this;
+            while (cur != null && cur!=this)
             {
                 if (cur.Data.Equals(data)) { return true; }
                 cur = cur.Next;
--- a/GenericsHomework/GenericsHomework/Node.cs
+++ b/GenericsHomework/GenericsHomework/Node.cs
@@ -1,41 +1,45 @@
-﻿using System;
-using System.Collections;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace GenericsHomework.GenericsHomework;
-
-public class Node<T>
+﻿namespace GenericsHomework
 {
-    public T? Data { get; set; }
-    private int Size { get; set; }
-    public Node<T> Next { get; set; }
-    public Node(T? data)
+    public class Node<T>
     {
-        Data = data;
-    }
-    public override string ToString()
-    {
-        return Data?.ToString() ?? "null";
-    }
-    public void Append(T data)
-    {
-        Next = new Node<T>(data);
-    }
-    public bool existing(T data)
-    {
-        if(data is null)
+        public T? Data { get; set; }
+        private int Size { get; set; }
+        public Node<T> Next { get; set; }
+        public Node(T? data)
         {
-            throw new ArgumentNullException(nameof(data));
+            Data = data;
         }
-        Node<T> cur = Next;
-        while(cur != null) 
+        public override string ToString()
         {
-            if (cur.Data.Equals(data)) { return true; }
-            cur = cur.Next;
+            return Data?.ToString() ?? "null";
         }
-        return false;
-
+        public void Append(T data)
+        {
+            Next = new Node<T>(data);
+        }
+        public void Clear()
+        {
+            Next = this;
+            /*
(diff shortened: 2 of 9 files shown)


BenFristadSchool (commented):
I will Review This

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            Data = data;
+        }
+        public override string ToString()
+        {
+            return Data?.ToString() ?? "null";
+        }
+        public void Append(T data)
+        {
+            Next = new Node<T>(data);
```
BenFristadSchool, line 18: Need to ensure that passed in value is not a duplicate of the linked list before adding

GenericsHomework/GenericsHomework/NodeTest.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.IsTrue(node.existing(20));
+            Assert.IsFalse(node.existing(40));
+        }
+
+        [TestMethod]
+        public void TestToStringMethod()
+        {
+            Node<int> node = new Node<int>(10);
+
+            Assert.AreEqual("10", node.ToString());
```
BenFristadSchool, line 51: Should change all ```Assert.AreEqual``` functions to compare based on type used in the test Example ```Assert.AreEqual<string>("10", node.ToString());```

GenericsHomework/GenericsHomework/NodeTest.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
```
BenFristadSchool, line 1: You should create a test project that references ```GenericsHomework.csproj``` rather than making a new class within the ```GenericsHomework.csproj```. Once you do that you will also need to add a ```Directory.Build.props``` file to consolidate the ```<PropertyGroup>``` xml in one file.

BenFristadSchool (commented):
Well Done!

arose-13:
Gonna review this

arose-13 (commented):
# Instructions
* Create a class library project called "GenericsHomework.". ✔
* Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ❌
* Define the Node class ✔
* The values should be homogeneous. ✔
* Include a constructor that takes a value. (No validation is necessary on the value). ✔
* Add a ToString() override that writes out the value's ToString() result. ✔
* Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ❌
* The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
* The Next property setter should be private. ❌
* Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔
* Whether it is sufficient to only set Next to itself ✔
* Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ✔
* Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ✔
* Create an Exists method to test to see if a value exists in the list. ✔
* Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ❌
* You should not rely on any BCL generic classes for your implementation. ✔
# Fundamentals
* Place all shared project properties into a Directory.Build.Props file.
* Place all shared project items into a Directory.Build.targets file.
* Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔
* Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔
* and enabled .NET analyzers for both projects ❌
* For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌
* All of the above should be unit tested ❌
* Choose simplicity over complexity ✔

I added comments for things that I noticed, but here's what I saw:
* There needs to be a separate project for the unit tests. I wasn't actually able to run the tests because of this.
* The linked list that you have built is not circular.
* `Append` method doesn't add to the list, it merely replaces the value of `Next` rather than inserting a new node.
* `Existing` method goes unused
* `Existing` also may loop indefinitely if the passed in data does not exist in the list
** To be fair though, since this list isn't circular the `Existing` method will stop eventually, but the list we are building is meant to be circular.
* Also be sure to use the generic version of `AreEqual`

Happy coding!

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,45 @@
...
+            */
+        }
+        public bool existing(T data)
+        {
+            if (data is null)
+            {
+                throw new ArgumentNullException(nameof(data));
+            }
+            Node<T> cur = Next;
+            while (cur != null)
```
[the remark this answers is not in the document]
Bhopss152: Thank you for the through review, fixed this.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+        public Node(T? data)
+        {
+
+            this.Data = data;
+            this.Next = this;
+        }
+        public override string ToString()
+        {
+            return Data?.ToString() ?? "null";
+        }
```
mmwoodfo, line 18:
```suggestion
        public override string? ToString()
        {
            return Data?.ToString();
        }
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
+﻿namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public T? Data { get; set; }
+        private int Size { get; set; }
+        public Node<T> Next { get; set; }
```
mmwoodfo, line 7:
```suggestion
        public Node<T> Next { get; private set; }
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+            Node<T> cur = this;
+            while (cur.Next != this)
+            {
+                cur = cur.Next;
+            }
+            Node<T> newNode = new Node<T>(data);
+            newNode.Next = this;
+            cur.Next = newNode;
+            this.Size++;
```
mmwoodfo, line 34: I would not recommend keeping track of size this way. I'd recommend adding a `Count` method to get the size of your list.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+            this.Next = this;
+        }
+        public override string ToString()
+        {
+            return Data?.ToString() ?? "null";
+        }
+        public void Append(T data)
+        {
+            if (data == null) return;
+            if(existing(data) == true)
```
mmwoodfo, line 22:
```suggestion
            if(existing(data))
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+            this.Data = data;
+            this.Next = this;
+        }
+        public override string ToString()
+        {
+            return Data?.ToString() ?? "null";
+        }
+        public void Append(T data)
+        {
+            if (data == null) return;
```
mmwoodfo, line 21:
```suggestion
            if (data is null) return;
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+            if(existing(data) == true)
+            {
+                throw new ArgumentException();
```
mmwoodfo, line 24:
```suggestion
                throw new ArgumentException("Cannot add duplicate item");
```
You should add some kind of message here to let the user know what's wrong, probably something better than what I wrote... But you get the idea. :)

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+            Next = this;
+            this.Size = 0;
+            /*
+            * The garbage collector in .NET automatically reclaims the memory
+            * occupied by unused nodes in a circular linked list. In class the
+            * professor talked about the automatic garbage collector .NET
+            * uses so that is where our answer is coming from.
+            */
+        }
+        public bool existing(T data)
```
mmwoodfo, line 47:
```suggestion
        public bool Exists(T data)
```

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace GenericsHomework
+{
+    [TestClass]
+    public class NodeTests
```
mmwoodfo, line 11: You should add a few more tests for testing your methods on different lengths of lists.

PR #114, opened 2023-02-13
Assignment 5 - Zach Lawson and Tom Rohr

zlawson-ewu:
Assignment 5 peer programmed by Zach Lawson (@zlawson-ewu) and Tom Rohr (@Tom-Rohr), no EC attempted.


joshmerrell:
I will review this.

joshmerrell:
It looks pretty good overall, and has very high test coverage.

## Instructions

  - Define the `Node` class  ✔

## Extra Credit

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file.  ✔
- Place all shared project items into a `Directory.Build.targets` file.  ❌

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+    public void Append(T value)
+    {
+        if (!Exists(value))
+        {
+            Node<T> temp = new(value) { Next = Next };
+            Next = temp;
+        }
+        else
+        {
+            throw new ArgumentException(nameof(value) + " already exists in the list.");
```
joshmerrell, line 24: Consider using an interpolated string here.

zlawson-ewu: A fair point. I did it this way out of habit really.

GenericsHomework/GenericsHomework.Test/NodeTests.cs
```diff
@@ -0,0 +1,93 @@
+namespace GenericsHomework.Test;
+
+[TestClass]
```
mmwoodfo, line 3: I would recommend adding a few more tests for testing your .Next property, you can check the assignment solution for some examples.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>
+{
+    public T? Value { get; set; } //Assuming "No validation is necessary on the value" means null is allowed.
```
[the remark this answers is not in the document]
zlawson-ewu: Ah I see, thank you for clearing that up.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+            temp = temp.Next;
+        } while (!temp.Equals(this) && !exists);
+        return exists;
+    }
+
+    public override string? ToString()
+    {
+        return (Value is null) ? "null" : Value!.ToString();
+    }
```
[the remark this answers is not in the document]
zlawson-ewu: Agreed, I intended to make this change after some clarification in Thursday's class, but it slipped my mind.

GenericsHomework/GenericsHomework.Test/NodeTests.cs
zlawson-ewu: Ah I see, we should've tried more list sizes. Thank you for the tip.

PR #115, opened 2023-02-14
Assignment5 @ lsobtafo

slanante:


BenagainB:
Define the Node class✔
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌✔
Implement System.Collections.Generic.ICollection<T> on the Node class ❌✔
Fundamentals

So, I'm still getting used to using generics, but you've used them very cleanly here. Especially like how you wrote your Dispose method inside Node.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        if (Content != null)
+        {
+            return Content.ToString()!;
+        }
+        else
+        {
+            return null;
+        }
+
```
mmwoodfo, line 25:
```suggestion
return Content.ToString();
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        if (Exists(content))
+        {
+            throw new Exception("The value within Content has already been appended previously.");
+        }
+
+        Node<T> newNode = new Node<T>(content);
+        Node<T> current = this;
+
+        if (NextNode.Equals(this))
```
mmwoodfo, line 38: You don't need to split this into an if/else statement, you should be able to just do something such as this:
```
           newNode.Next = Next;
            Next = newNode;
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,116 @@
...
+        List<Node<T>> nodes = new List<Node<T>>();
+
+        while (currNode.NextNode != currNode)
+        {
+            currNode = currNode.NextNode;
+            nodes.Add(currNode);
+        }
+        foreach (Node<T> node in nodes)
+        {
+            Dispose(node);
```
mmwoodfo, line 83: You do not need to call dispose here

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+        NextNode = this;
+        PrevNode = null;
+    }
+
+    public bool Exists(T content)
+    {
+        Node<T> currNode = this;
+        while (currNode != null)
```
mmwoodfo, line 92: When will current node equal null?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    Or at least try to...
+
+    */
+    public void Clear()
+    {
+        Node<T> currNode = this;
+        List<Node<T>> nodes = new List<Node<T>>();
+
+        while (currNode.NextNode != currNode)
```
mmwoodfo, line 76: You want to stop your loop when you reach the end of your list, you'll reach the end of your list when your `Next` property references back to it's self.

GenericsHomeWork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,45 @@
+namespace GenericsHomework.Tests;
+
+[TestClass]
+public class NodeTests
+{
+    [TestMethod]
+    public void Node_Append_Success()
```
mmwoodfo, line 7: You should have a test for to string in here as well as some more tests for testing clearing, appending and traversing on different list lengths.

PR #117, opened 2023-02-14
Assignment 5 - Andrew Korchemniy and Vera Korchemnaya

AndrewKorchemniy:


hallen10:
I will review this.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public T Value { get; set; }
+    public Node<T> Next
+    {
+        get => _Next!;
+        private set => _Next = value ?? throw new ArgumentNullException(nameof(value));
+    }
+
+    private Node<T>? _Next;
```
hallen10, line 19: A warning here says that: 
"Name '_Next' does not match rule 'Instance fields (private)'. Suggested name is '_next'".

hallen10 (commented):
## Instructions

  - Define the `Node` class

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌✔

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌✔

## Fundamentals

JessicaH46:
Define the Node class

Your code looks good! Looks like you got all the components to the assignment and the fundamentals are updated! All the comments explaining your logic were very easy to understand and read, I think you guys did really good! Code was very organized and easy to read, great job!

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,103 @@
+using System;
+namespace GenericsHomeworkTests;
+
+[TestClass]
+public class NodeTests
```
VeraKorchemnaya, line 5: Note to self: rearrange tests to reflect the Node class structure.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+namespace GenericsHomework;
+
+public class Node<T> where T : struct
```
VeraKorchemnaya, line 4: Note to self: the constraint should probably be notnull instead of struct

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+namespace GenericsHomework;
+
+public class Node<TValue>
+{
+    public TValue? Value { get; set; }
+    private Node<TValue>? _Next;
+    public Node<TValue> Next { get => _Next!; private set => _Next = value ?? throw new ArgumentNullException(nameof(value)); }
```
mmwoodfo, line 8: I would suggest multi-lining this

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,50 @@
...
+            iterator = iterator.Next;
+        }
+        return false;
+    }
+
+    public override string ToString()
+    {
+        if (Value is null) return string.Empty;
+        else return Value.ToString() ?? string.Empty;
+    }
```
mmwoodfo, line 49:
```suggestion
    public override string ToString()
    {
        return Value?.ToString() ?? string.Empty;
    }
```
You could also change the method signature, there is a nullable override option. Then you can just return null.
`public override string? ToString()`

mmwoodfo (commented):
Great job!

GenericsHomework/GenericsHomework/Node.cs
VeraKorchemnaya: We appreciate your feedback, thank you!

PR #118, opened 2023-02-14
Assignment5 - Nolan Posey & Josh Merrell (Extra credit attempted)

Nolan-Posey:
@Nolan-Posey
@joshmerrell

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -23,15 +23,17 @@ public class NodeTests
     public void Add_AddsNewValue()
     {
         //Arrange
-        Node<int> head = new(1, true);
+        CircularSinglyLinkedList<int> list = new();
+        Node<int> head;
 
         //Act
-        head.Add(42);
+        list.Append(1);
+        head = list.Head!;
+        head.Add(2);
 
         //Assert
         Assert.AreEqual<int>(2, head.Count);
-        Assert.IsFalse(head.Next.IsHead);
-        Assert.AreEqual<int>(42, head.Next.Data);
+        Assert.AreEqual<int>(2, head.Next.Data);
     }
 
 
@@ -40,9 +42,12 @@ public class NodeTests
     {
         //Arrange
         int[] ints = {0, 0, 0, 0, 0};
-        Node<int> head = new(1, true);
+        CircularSinglyLinkedList<int> list = new();
+        Node<int> head;
 
         //Act
+        list.Append(1);
+        head = list.Head!;
         head.Add(2);
         head.Add(3);
         head.Add(4);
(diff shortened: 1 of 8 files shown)


StrangeRanger:
Will review

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,151 @@
...
+    private T? _Data;
+
+    public Node<T> Next
+    {
+        get => _Next!;
+        private set => _Next = value ?? throw new ArgumentNullException(nameof(Next));
+    }
+
+    private int _Count;
+    public int Count => _Count!;
```
StrangeRanger, line 21:
```suggestion
    public int Count => _Count;
```
IDE Message: "The nullable warning suppression expression is redundant"

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        CircularSinglyLinkedList<string> list = new();
+        Node<string> head;
+        IEnumerator<string> headEnumerator;
+        string[] iteratorComparison = { "one", "two", "three", "four" };
+        int index = 0;
+
+        //Act
+        list.Append("one");
+        head = list.Head!;
+        headEnumerator = head.GetEnumerator();
```
StrangeRanger, line 77: IDE Warning: "Instance of IEnumerator is never disposed"

StrangeRanger (commented):
## Instructions

  - Define the `Node` class
  - Add a `ToString()` override that writes out the value's `ToString()` result. ❌
  - Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ❌

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ✔

## Fundamentals

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,151 @@
...
+    public bool IsReadOnly => false;
+
+    private CircularSinglyLinkedList<T>? _List;
+    protected CircularSinglyLinkedList<T> List
+    {
+        get => _List!;
+        init => _List = value ?? throw new ArgumentNullException(nameof(List));
+    }
+
+    private Node<T>? _Next;
```
mmwoodfo, line 32: I'd suggest moving this closer to your getter/setter property

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,151 @@
...
+    public Node(T data)
+    {
+        Data = data;
+        Next = this;
+        _Count = 1;
```
mmwoodfo, line 37: I'd recommend making a method that traverses your list and counts, rather than trying to track it this way. It's easier to make mistakes doing it this way.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,151 @@
...
+        Node<T> newNode = new(data, this.List);
+        temp.Next = newNode;
+        newNode.Next = head;
+        _Count++;
+    }
+
+    public override string ToString()
+    {
+        return Data!.ToString()!;
+    }
```
mmwoodfo, line 61:
```suggestion
    public override string? ToString()
    {
        return Data?.ToString();
    }

```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    public Node(T data, CircularSinglyLinkedList<T> list) : this(data)
+    {
+        List = list;
+    }
+
+    public void Append(Node<T> head, T data)
+    {
+        Node<T> temp = head;
```
mmwoodfo, line 47: You should check if the data is duplicated here

mmwoodfo (commented):
You had alot of good comments from peers, make sure you go through those comments.

PR #119, opened 2023-02-14
Assignment5: @crawfordkid2 and @GeorgeKoralski007

GeorgeKoralski007:


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>
+{
+    public T Value { get; set; }
```
jdriscoll1, line 5: Make this property nullable

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+            Node<T> next = current.Next;
+            current.Next = current;
+            current = next;
+        } while (current != this);
+    }
+
+    public void Append(T value)
+    {
+        if (Exists(value))
+            throw new ArgumentException("Value already exists in the list", nameof(value));
```
jdriscoll1, line 33: Looks great, although changing it to a format string such as: `ArgumentException($"Value already exists in the list {nameof(value)}")` or instead putting a space after "list" might make it neater.

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void Node_Constuctor_Success()
+        {
+            Node<int> node = new Node<int>(10);
```
jdriscoll1, line 12: Change to `Node<int> node = new(10);`

jdriscoll1 (commented):
Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ✔
Define the Node class ✔
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌

Provide a comment & a test that shows why you chose to go through each node and set its next to itself
See above
Extra Credit
Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌
Implement System.Collections.Generic.ICollection<T> on the Node class ❌
Fundamentals
Place all shared project items into a Directory.Build.targets file. ✔
Not enabled on GenericsHomeworkTests.cproj
Lang version not set on tests
and enabled .NET analyzers for both projects ❌
EnableNETAnalyzers
Add <type> on AreEqual tests in unit tests
Add unit test to display why you chose to do clear and you'll be good. 

Overall it looks great, simple and clean design. I would recommend that you add comment and a test displaying why you chose to go through each value and close the loop in the clear method, and make a few tweaks.

GenericsHomework/GenericsHomework/Node.cs
GeorgeKoralski007: Done

GenericsHomework/GenericsHomework/Node.cs
GeorgeKoralski007: Added a space

GenericsHomework/GenericsHomeworkTests/NodeTests.cs
GeorgeKoralski007: Done

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,143 @@
...
+    public Node(T? value)
+    {
+        Value = value;
+        Next = this;
+    }
+
+    public override string ToString()
+    {
+        return Value?.ToString() ?? "null";
+    }
```
mmwoodfo, line 33:
```suggestion
    public override string? ToString()
    {
        return Value?.ToString();
    }

```
Could also do this

mmwoodfo (commented):
Great job!

PR #120, opened 2023-02-14
Assignment 5 - Hunter Thompson & Hunter Allen

StrangeRanger:
@hallen10


Malcolm-Shepherd:
## Instructions

  - Define the `Node` class✔
  - The values should be homogeneous.✔
  - Add a `ToString()` override that writes out the value's `ToString()` result. ❌
    - The `Next` property should be non-nullable (careful to follow the non-nullable property guidelines) ❌
  - Add an `Append` method that takes a value and appends a new `Node` instance after the current node (by invoking the `Next` property). ❌
## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file.✔
- Place all shared project items into a `Directory.Build.targets` file.✔

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+    public void Append(T value)
+    {
+        if (Exists(value))
+        {
+            throw new DuplicateDataInArrayException(nameof(value));
+        }
+
+        Node<T> current = this;
+
+        while (current.Next != current) { current = current.Next; }
```
Malcolm-Shepherd, line 24: This Appends the new Node at the end of the list, not after the current node

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+            count++;
+        }
+
+        return count;
+    }
+    */
+
+    public override string ToString()
+    {
+        string holder = "{ ";
```
JessicaH46, line 72: For your ToString, I would simplify the code to something similar to this
`if(Value != null) return Value.ToString();
            else return string.Empty;`
this is what we had for ours and it did the job!

JessicaH46:
Define the Node class
Add a ToString() override that writes out the value's ToString() result. ❌✔ * Could simplify, left a comment
Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ❌✔

Overall you have all the main components! The fundamentals all look up to date and your tests all look good to me! Great job! I did leave a comment on your toString with a suggestion, I also like how you checked to see if the data was laready in the list with your append! I didn't think to do it that way, I saw someone else's comment on your append, but ours looked pretty similar except for at the end of your loop we had an extra line that would be 
current.Next.Next = this;
Not sure if that's what would resolve your problem, but it might so I will leave it here. Good job on the assignment!

GenericsHomework/Node.cs
StrangeRanger: Is that not the point? If I do it after the current node, it'd be constantly overriding the second node in the list.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+        Node<T> current = this;
+
+        while (current.Next != current)
+        {
+            holder += current.Value + ", ";
+            current = current.Next;
+        }
+
+        holder += current.Value + " }";
+        return holder;
```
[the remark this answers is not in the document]
StrangeRanger: Thank you for your suggestion here.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+public class Node<T>
+{
+    // Must be non-nullable.
+    public Node(T value)
+    {
+        Value = value !;
+        Next = this;
+    }
+
+    public Node<T> Next { get; private set; }
```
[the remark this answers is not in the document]
StrangeRanger: Thanks for pointing that out

GenericsHomework/Node.cs
StrangeRanger: That's a good idea to have it check if the value is null. I implemented it as such: `return Value is not null ? Value.ToString()! : string.Empty;`

mmwoodfo (commented):
Good job!

PR #121, opened 2023-02-14
Assignment5 @BenFristadSchool and @mmoran314

BenFristadSchool:


Bhopss152 (commented):
Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. **X**
Define the Node class ✔
The values should be homogeneous.✔
The Next property setter should be private. **X**
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). **X**
Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) **X**


 
Place all shared project properties into a Directory.Build.Props file. X
Place all shared project items into a Directory.Build.targets file. X

Good code, I went trough the checklist and there are some small things that your missing like making something private instead of public, clear method, and proj file.

Tom-Rohr:
I'll review this.

Tom-Rohr (commented):
  - Define the `Node` class
-- Place all shared project properties into a `Directory.Build.Props` file.❌
- Place all shared project items into a `Directory.Build.targets` file. ❌
- Choose simplicity over complexity ❌✔

Nice Job. Everything works to assignment specifications, but you still need to consolidate shared properties into a Directory.Build.Props file. My suggestions don't change the functionality of your code at all, just little clean up things.

GenericsHomework/GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+	  <TargetFramework>net7.0</TargetFramework>
+	  <ImplicitUsings>enable</ImplicitUsings>
+	  <Nullable>enable</Nullable>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <LangVersion>11.0</LangVersion>
```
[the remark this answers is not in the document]
BenFristadSchool: I tried to adding the ```Directory.Build.props``` file to the outer GenericsHomework Directory and moved the ```<Project>``` and ```<PropertyGroup>``` xml along with all the tags within ```<PropertyGroup>``` but I could not get the project to build after. I was researching for a couple hours but couldn't find a solution.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            Assert.AreEqual<Node<int>>(node, node.Next.Next);
+            Assert.AreEqual<int>(node.Next.Value, 2);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentException))]
+        public void Append_Failure()
+        {
+            node.Append(2);
+            node.Append(2);
```
mmwoodfo, line 39: This test doesn't seem to be passing

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,60 @@
...
+         */
+        public void Clear()
+        {
+            Next = this;
+        }
+
+        public bool Exists(T value)
+        {
+            bool result = false;
+            Node<T> curr = Next;
```
mmwoodfo, line 45:
```suggestion
            Node<T> curr = this;
```
This will get your test passing

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public override string ToString()
+        {
+            return Value!.ToString() ?? "value is null";
+        }
+
+        public void Append(T newValue)
+        {
+            Next = !Next.Exists(newValue) ? new Node<T>(newValue, Next) : throw new ArgumentException("value already exists in linked list");
```
mmwoodfo, line 21: It might be more readable if you make this multiple lines

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public class Node<T>
+    {
+        public T Value { get; set; }
+        public Node<T> Next { get; private set; }   
+
+        public Node(T value, Node<T> next)
+        {
+            Value = value;
+            Next = next ?? this;
```
mmwoodfo, line 11: I would recommend using the append method to handle this, rather than passing the next option into the constructor.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public override string ToString()
+        {
+            return Value!.ToString() ?? "value is null";
+        }
```
mmwoodfo, line 17:
```suggestion
        public override string? ToString()
        {
            return Value?.ToString();
        }
```
Another option

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,68 @@
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class NodeTests
```
mmwoodfo, line 4: This could probably use a few more tests to test the various methods with different list lengths.

PR #122, opened 2023-02-14
Assignment5 alan and grant

Howard404:
Done by @Howard404 and @arose-13

### DIFF
--- a/GenericsHomework.Tests/NodeTest.cs
+++ b/GenericsHomework.Tests/NodeTest.cs
@@ -1,11 +1,10 @@
-using Microsoft.VisualStudio.TestTools.UnitTesting;
 namespace GenericsHomework.Tests;
 
 [TestClass]
 public class NodeTest
 {
     [TestMethod]
-    public void CreateNode_ContructorTakesMultipleTypes_ReturnsNotNull()
+    public void CreateNode_ConstructorTakesMultipleTypes_ReturnsNotNull()
     {
         //Assemble
         Node<int> node1 = new(30);
@@ -31,9 +30,9 @@ public class NodeTest
         //Act
 
         //Assert
-        Assert.AreEqual<string>("Node Value: 30", node1.ToString());
-        Assert.AreEqual<string>("Node Value: Some Value", node2.ToString());
-        Assert.AreEqual<string>("Node Value: 3.14", node3.ToString());
+        Assert.AreEqual<string>("Node value: 30", node1.ToString());
+        Assert.AreEqual<string>("Node value: Some Value", node2.ToString());
+        Assert.AreEqual<string>("Node value: 3.14", node3.ToString());
     }
 
     [TestMethod]
@@ -49,7 +48,7 @@ public class NodeTest
     }
 
     [TestMethod]
-    public void GivenData_AppendsNodeToExistingNode()
+    public void Appends_GivenData_AppendsNodeToExistingNode()
     {
         // Assemble
         Node<int> node1 = new(12);
@@ -90,6 +89,21 @@ public class NodeTest
 
         // Assert
         Assert.IsTrue(node1.Exists(node3.Data));
+    }
+
+    [TestMethod]
+    public void Exists_IfNodeAlreadyExists_ReturnFalse()
+    {
+        // Assemble
+        Node<int> node1 = new(100);
+        Node<int> node2 = new(200);
+        Node<int> node3 = new(200);
+        Node<int> node4 = new(300);
+
+        // Act
+        node1.Append(node2);
+
+        // Assert
         Assert.IsFalse(node1.Exists(node4.Data));
     }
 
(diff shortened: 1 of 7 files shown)


jacobnewcomb (commented):
The only issue I see with your code is that the append method doesn't work like it is supposed to. In your implementation, Next (for the head node) is constantly being changed to the newest node added to the list, which isn't traditionally how an append method for a list works. Typically, the newest node would be appended to the end of the list.

VeraKorchemnaya:
I will review this.

GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,135 @@
...
+        Node<int> node3 = new(200);
+        Node<int> node4 = new(300);
+
+        // Act
+        node1.Append(node2);
+
+        // Assert
+        Assert.IsTrue(node1.Exists(node3.Data));
+        Assert.IsFalse(node1.Exists(node4.Data));
+    }
```
VeraKorchemnaya, line 94: In this method you seem to be testing two things but the title doesn't reflect that. You make two assertions, 1. That if a value exists the Exists method will return true and 2. That if a value doesn't exist it will return false. Consider splitting into two tests.

GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    [TestMethod]
+    [ExpectedException(typeof(InvalidDataException))]
+    public void Append_IfDataExists_ThrowError()
+    {
+        // Asseble
```
VeraKorchemnaya, line 100: You misspelled Assemble here and in the method below. You can look at your error list messages after you build to catch these kinds of little typos.

GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,135 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace GenericsHomework.Tests;
+
+[TestClass]
+public class NodeTest
+{
+    [TestMethod]
+    public void CreateNode_ContructorTakesMultipleTypes_ReturnsNotNull()
```
VeraKorchemnaya, line 8: Small typo here on Constructor.

VeraKorchemnaya (commented):
Instructions
Create a class library project called "GenericsHomework.". ✔ (I would recommend renaming your solution file name to make the project name)
Define the Node class
The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ❌ (See comment)
Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔ (Technically works but I would recommend looking at it some more and refactoring it. There is some extra code there)
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌ (If you are implementing it this way you need to show why it is necessary)
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌ (You did not provide a comment about this)

Fundamentals
Place all shared project properties into a Directory.Build.Props file. ❌ (You didn't do this)
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional)   ✔
All of the above should be unit tested ✔ (Overall pretty good, just take a look at my comment above about testing the clear method)

Overall you did a great job!

VeraKorchemnaya:
>

VeraKorchemnaya:
> The only issue I see with your code is that the append method doesn't work like it is supposed to. In your implementation, Next (for the head node) is constantly being changed to the newest node added to the list, which isn't traditionally how an append method for a list works. Typically, the newest node would be appended to the end of the list.

This is not correct because we don't have a head node here. When you call append you are changing the Next of the node that is calling append which is exactly what you want. There also isn't an "end" here because it is a circularly linked list. I totally understand why you would come to that conclusion though because I too was struggling with the concept of this data structure even though we've done circularly linked lists before in other classes. It really helped me when I drew it out on paper and stepped through the code.

arose-13:
> The only issue I see with your code is that the append method doesn't work like it is supposed to. In your implementation, Next (for the head node) is constantly being changed to the newest node added to the list, which isn't traditionally how an append method for a list works. Typically, the newest node would be appended to the end of the list.

Normally I would agree, but the instructions state that we should append the "new Node instance after the current node". Since we assign the new node's `Next` value to the current node's `Next` value, we can keep the rest of the list intact (even though the way this method works is more of an `Insert` method). Thank you for your feedback!

arose-13:
Thanks for the feedback guys!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        Node<T> newNode = new(data);
+        newNode.Next = Next;
+        Next = newNode;
+
+    }
+
+    public void Append(Node<T> newNode)
+    {
+        if (Exists(newNode.Data)) throw new InvalidDataException($"The specified data {nameof(newNode)} already exists in the list!");
```
mmwoodfo, line 32: An argument exception would be an acceptable exception to throw here

mmwoodfo (commented):
Good job!

PR #123, opened 2023-02-14
Assignment5 1bbartels

1bbartels:


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public T Value;
+        public Node<T> Next { get; private set; } 
+
```
lsobtafo, line 7: Try to use properties instead of public fields for Value and Next. Try using properties with a private set method to control how the fields are accessed and modified.

Afterlyte (changes requested):
## Instructions

  - Define the `Node` class
  - Add a `Next` property that references the next node or else refers back to itself if there are no other nodes in the list. ❌
    - **This should be modified so that the property is doing the referencing back to itself instead of managing that in other places of the code**
    - The `Next` property should be non-nullable (careful to follow the non-nullable property guidelines) ❌
     - **Similar note as above**
    - Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌
     - **Only functionality right now is setting Next to itself**
      - **Same as above note, no other functionality is there**

## Extra Credit

## Fundamentals

  - **AreEqual was never used in the first place (you probably should since it's useful for more precise data validation rather than just using IsTrue on the Exists method**

Overall well done on the main fundamentals, main things to work on are more precise unit tests and some bits and pieces of code in the primary functionalities of the project (like the properties, ToString, Append, etc.

GenericsHomework.Tests/Node.Tests.cs
```diff
@@ -0,0 +1,50 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
```
mmwoodfo, line 6: It would be good to have some more tests that test various lengths of your list against your different methods.

mmwoodfo (commented):
There were alot of good comments left by your peers on this PR.

PR #124, opened 2023-02-14
Assignment5

Clockknight:
Partnered with @Malcolm-Shepherd 

Commiting with base requirements completed for now, will put in EC later.

### DIFF
--- a/ClassLibraryProjects/GenericsHomework/GenericsHomework.csproj
+++ b/ClassLibraryProjects/GenericsHomework/GenericsHomework.csproj
@@ -1,9 +1,3 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
-  </PropertyGroup>
-
 </Project>
--- a/ClassLibraryProjects/GenericsHomework/GenericsHomework.csproj
+++ b/ClassLibraryProjects/GenericsHomework/GenericsHomework.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+</Project>
--- a/ClassLibraryProjects/GenericsHomework/Node.cs
+++ b/ClassLibraryProjects/GenericsHomework/Node.cs
@@ -1,0 +1,7 @@
+﻿namespace GenericsHomework
+{
+    public class Class1
+    {
+
+    }
+}
--- a/ClassLibraryProjects/GenericsHomework/Node.cs
+++ b/ClassLibraryProjects/GenericsHomework/Node.cs
@@ -2,15 +2,36 @@
 {
     public class Node<T>
     {
+        private Node<T>? _Next;
         public T Value { get; set; }
+        
+        // I don't know how to unit test for null with the way Next is being used. //
+        public Node<T> Next
+        {
+            get
+            {
+                return _Next!;
+            }
+            private set
+            { 
+                _Next = value ?? throw new ArgumentNullException(nameof(value));
+            }
+        }
         public Node(T value) 
         { 
             Value = value;
+            Next = this;
         }
 
         public override string ToString() 
         {
             return this.Value!.ToString()!;
+        }
+
+        public void Append(Node<T> nextNode) 
+        { 
+            nextNode.Next = this.Next;
+            this.Next = nextNode;    
         }
     }
 }
--- a/ClassLibraryProjects/GenericsHomework/Node.cs
+++ b/ClassLibraryProjects/GenericsHomework/Node.cs
@@ -6,7 +6,7 @@ namespace GenericsHomework
     public class Node<T>
     {
         private Node<T>? _next;
-        public T? Value { get; set; }
+        public T Value { get; set; }
         
         // I don't know how to unit test for null with the way Next is being used. //
         public Node<T> Next
@@ -20,7 +20,7 @@ namespace GenericsHomework
                 _next = value ?? throw new ArgumentNullException(nameof(value));
             }
         }
-        public Node(T? value) 
+        public Node(T value) 
         { 
             Value = value;
             Next = this;
--- a/ClassLibraryProjects/GenericsHomework/Node.cs
+++ b/ClassLibraryProjects/GenericsHomework/Node.cs
@@ -1,4 +1,7 @@
-﻿namespace GenericsHomework
+﻿using System.Data;
+using System.Runtime.InteropServices;
+
+namespace GenericsHomework
 {
     public class Node<T>
     {
@@ -30,8 +33,44 @@
 
         public void Append(Node<T> nextNode) 
         { 
+            if (this.Exists(nextNode.Value))
+            {
+                throw new ArgumentException("Duplicate Value detected");
+            }
+
             nextNode.Next = this.Next;
             this.Next = nextNode;    
+        }
+
+        public void Clear()
+        {
+            Node<T> cur = this.Next;
+            Node<T> temp = cur;
+            while (cur != this)
+            {
+                cur = cur.Next;
+                temp.Next = temp;
+                temp = cur;
+            }
+            this.Next = cur;
+
+            // There is no need to collect garbage. C# .Net automatically collects garbage on memory space that is no longer used
+        }
+
+        public bool Exists(T value)
+        {
+            Node<T> cur = this;
+
+            do
+            {
+                if (value.Equals(cur.Value))
+                { return true; }
+                cur = cur.Next;
+            }
+            while (cur != this);
+
+            return false;
+
         }
     }
 }
--- a/ClassLibraryProjects/GenericsHomework/Node.cs
+++ b/ClassLibraryProjects/GenericsHomework/Node.cs
@@ -1,7 +1,16 @@
 ﻿namespace GenericsHomework
 {
-    public class Class1
+    public class Node<T>
     {
+        public T Value { get; set; }
+        public Node(T value) 
+        { 
+            Value = value;
+        }
 
+        public override string ToString() 
+        {
+            return this.Value!.ToString()!;
+        }
     }
 }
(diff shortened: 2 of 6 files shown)


jstanley6:
I will review

jstanley6:
Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ✔
Define the Node class

Fundamentals
Place all shared project properties into a Directory.Build.Props file. ❌
Place all shared project items into a Directory.Build.targets file. ❌
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) 
and enabled .NET analyzers for both projects ❌

ClassLibraryProjects/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -47,4 +47,87 @@ public void Append_Appends_Pass()
...
+        Node<int> node1 = new(42);
+        Node<int> node2 = new(0);
+        Node<int> node3 = new(314);
+        node1.Append(node2);
+        node2.Append(node3);
+
+
+        Assert.IsTrue(node1.Exists(0));
+    }
+
```
jstanley6, line 94: forgot to add the [TestMethod] above this unit test

jstanley6 (commented):
I can't find the Directory.Build.props file

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -2,15 +2,36 @@
...
+        public Node<T> Next
+        {
+            get
+            {
+                return _Next!;
+            }
+            private set
+            { 
+                _Next = value ?? throw new ArgumentNullException(nameof(value));
+            }
```
jstanley6, line 21: instead of what you got for Next, you can do this instead
public Node<T> Next
        {
            get => _Next ?? this;
            private set => _Next = value;
        }
Also the naming convention for _Next is wrong, it should be _next

jstanley6:
Other then the few things I mentioned everything else looks good.

AndrewKorchemniy:
Will review

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        public void Append(Node<T> nextNode) 
+        { 
+            if (this.Exists(nextNode.Value))
+            {
+                throw new ArgumentException("Duplicate Value detected");
+            }
+
+            nextNode.Next = this.Next;
+            this.Next = nextNode;    
+        }
```
AndrewKorchemniy, line 43:
```suggestion
        public void Append(T value) 
        { 
            if (Exists(value))
                throw new ArgumentException("Duplicate Value detected");
            Node<T> newNode = new(value);
            newNode.Next = Next;
            Next = newNode;    
        }
```
Append should take in a T value, not a Node of type T. Using the "this" keyword here is also not necessary. You could argue it improves readability, but the IDE recommends simplification. This will break a lot of tests. Just have the tests pass in the values instead of actual nodes.

ClassLibraryProjects/GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,133 @@
...
+    {
+        Node<int> node1 = new(42);
+        Node<int> node2 = new(0);
+        Node<int> node3 = new(314);
+        node1.Append(node2);
+        node2.Append(node3);
+
+        Assert.IsFalse(node1.Exists(1337));
+
+    }
```
AndrewKorchemniy, line 117:
```suggestion
    public void Exists_NonExistingValue_ReturnsFalse()
    {
        // Arrange
        Node<int> node = new(42);
        node.Append(int.MaxValue);
        
        // Assert
        Assert.IsFalse(node.Exists(0));
    }
```
An example of how to make the tests work again after accepting suggestions.

ClassLibraryProjects/GenericsHomework/Node.cs
AndrewKorchemniy: I agree actually, since the standard c# convention is _camelCase. However, I believe our prof prefers _PascalCase.

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+            Node<T> temp = cur;
+            while (cur != this)
+            {
+                cur = cur.Next;
+                temp.Next = temp;
+                temp = cur;
+            }
+            this.Next = cur;
+
+            // There is no need to collect garbage. C# .Net automatically collects garbage on memory space that is no longer used
```
AndrewKorchemniy, line 57: I doesn't look like the questions were really addressed here. See the above comment for an explaination.
https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals

The questions:
Whether it is sufficient to only set Next to itself

Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required).

Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected.

AndrewKorchemniy (commented):
  - Define the `Node` class
  - Add an `Append` method that takes a value and appends a new `Node` instance after the current node (by invoking the `Next` property). ❌
    - Whether it is sufficient to only set Next to itself ❌
    - Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌
- Ensure that you turn on code analysis for all projects(EnableNETAnalyzers)  ❌
- Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional)   ❌
- and enabled .NET analyzers for both projects ❌
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version) ✔

Extra credit was not attempted
There was one warning

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Data;
+using System.Runtime.InteropServices;
+
+namespace GenericsHomework
+{
+    public class Node<T>
```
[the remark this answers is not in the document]
AndrewKorchemniy: Honestly, I'm not too sure if we should limit it to value type or not. The question asked in the teams chat kind of confused me a bit. You could just set value to be nullable and add logic to handle it - instead of limiting it to value types. The instructions do say "value" a lot though. Something to ask about in-class.

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                if (value.Equals(cur.Value))
+                { return true; }
+                cur = cur.Next;
+            }
+            while (cur != this);
+
+            return false;
+
+        }
```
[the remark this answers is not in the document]
AndrewKorchemniy: Scratch that. If you make this.Value nullable, just use string.Equals() to compare the values. That way you're not relying on a potentially null value to access equals().

ClassLibraryProjects/GenericsHomeworkTests/NodeTests.cs
Malcolm-Shepherd: Fixed

ClassLibraryProjects/GenericsHomework/Node.cs
Malcolm-Shepherd: Since they want Next to be non-nullable, checking the value on the setter is important. I do agree with the field name change though.

ClassLibraryProjects/GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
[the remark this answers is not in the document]
Malcolm-Shepherd: Refactored into Directory.Build.Props

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        public Node(T value) 
+        { 
+            Value = value;
+            Next = this;
+        }
+
+        public override string ToString() 
+        {
+            return this.Value!.ToString()!;
```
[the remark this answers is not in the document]
Malcolm-Shepherd: This is definitely better than the way I did it.

ClassLibraryProjects/GenericsHomework/Node.cs
Malcolm-Shepherd: I think after this was clarified in class, this seems fine.

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            Node<T> cur = this.Next;
+            Node<T> temp = cur;
+            while (cur != this)
+            {
+                cur = cur.Next;
+                temp.Next = temp;
+                temp = cur;
+            }
+            this.Next = cur;
```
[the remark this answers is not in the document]
Malcolm-Shepherd: Simplified and added a more clear comment.

ClassLibraryProjects/GenericsHomework/Node.cs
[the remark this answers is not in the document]
Malcolm-Shepherd: I think your first solution is better. Making it nullable allows the value to be null, which does not change anything here.

ClassLibraryProjects/GenericsHomework/Node.cs
Malcolm-Shepherd: Fixed Append and the tested it would break

ClassLibraryProjects/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,71 @@
...
+        public Node(T value) 
+        { 
+            Value = value;
+            Next = this;
+        }
+
+        public override string ToString() 
+        {
+            if (Value is null) return string.Empty;
+            else return Value.ToString() ?? string.Empty;
```
mmwoodfo, line 32:
```suggestion
            return Value?.ToString() ?? string.Empty;
```

ClassLibraryProjects/GenericsHomework/Node.cs
mmwoodfo: Yes, we do prefer pascal. However, I will never deduct points for using camel case, since it is c# convention. I will just assume your intent was to follow c# convention, not that you forgot we use pascal in the class. :)

mmwoodfo (commented):
image
This should be a class library, not a console app

PR #125, opened 2023-02-14
Assignment5 @BenagainB

mhtran23:


slanante (commented):
Overall great job, did not find much other than small possible additions to the code. The Fundamentals portion of homework description about the Directory.Build.props file was not implemented. But again, great job I really liked the constructor that starts in line 16 that allows for appending nodes without using Append.

Define the Node class
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌

Fundamentals

Clockknight:
Define the Node class
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌

Fundamentals
For this assignment, always use Assert.AreEqual() (the generic version) ✔

I didn't see any comments on thoughts behind the considerations provided for Node's clear() behavior, but to be fair I also skimmed on those in my PR, haha.

Pretty good for the most part in terms of functionality, seems good, though.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,79 @@
+﻿namespace GenericsHomework
+{
+    public class Node<TValues> where TValues : notnull
+    {
+        public TValues Value { get; set; }
+        public Node<TValues> Next { get; private set; }
+        public Node<TValues> Last;
```
mmwoodfo, line 7: You may consider making this also a private set. You also should put a getter on this if you intend for it to be a property rather than a field.

mmwoodfo, line 7: Whats the reasoning behind adding a last? As this isn't part of the assignment and does complicate it slightly.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            Value = value;
+            Next = node;
+            Last = this;
+        }
+
+        public void Append(TValues value)
+        {
+            bool keyValueExists = Exists(value);
+            if (keyValueExists == true)
```
mmwoodfo, line 26:
```suggestion
            if (Exists(value))
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,79 @@
...
+        public Node(TValues value, Node<TValues> node)
+        {
+            Value = value;
+            Next = node;
+            Last = this;
+        }
+
+        public void Append(TValues value)
+        {
+            bool keyValueExists = Exists(value);
```
mmwoodfo, line 25:
```suggestion
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public override string ToString()
+        {
+            string? stringReturn = null;
+            if (Value is not null)
+                stringReturn = Value.ToString();
+            if (stringReturn is null)
+                return "";
+            return stringReturn;
+        }
```
mmwoodfo, line 77:
```suggestion
        public override string? ToString()
        {
            return Value?.ToString();
        }
```

PR #126, opened 2023-02-14
Assignment5

DaisyPalencia:
Pairing with @AnnikaStachowiak


jacobnewcomb (commented):
- Create a class library project called "GenericsHomework.". ✔
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.✔
- Define the Node class✔
- The values should be homogeneous.✔
- Include a constructor that takes a value. (No validation is necessary on the value). ✔
- Add a ToString() override that writes out the value's ToString() result. ✔
- Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
- The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
- The Next property setter should be private. ✔
- Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔

Everything looks great! I really like how you did your clear method, I think I will implement this strategy in my own fork!

GenericsHomework/GenericsHomework.Tests/GenericsHomeworkTests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
alowe4, line 1: why are there two test csproj files????

GenericsHomework/GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
```
alowe4, line 7: needs EnableNETAnalyzers, and LangVersion 11

alowe4 (commented):
 Define the Node class✔
 The values should be homogeneous.✔	
Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following:✔
 Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌[no comment exists for this]
Create an Exists method to test to see if a value exists in the list. ✔[you made one, it just doesn't work]


Fundamentals

Place all shared project items into a Directory.Build.targets file.✔

---------------------------------------------------------------------------
main things to fix 
 -the lack of a Directory.Build.Props file,
 -the 2 test csproj files
 -the lack of testing on append and exists
 -missing comment about garbage collection

GenericsHomework/GenericsHomework.Tests/GenericsHomeworkTests.csproj
DaisyPalencia: I think this is just a mistake? It doesn't show up when I look for it.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System.Runtime.CompilerServices;
+
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public T Value { get; }
+
+        public Node<T> Next { get; private set; }
```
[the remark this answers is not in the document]
DaisyPalencia: We wanted to have it use the "?" so the code would be more understandable and clearer.

GenericsHomework/GenericsHomework.Tests/GenericsHomework.Tests.csproj
DaisyPalencia: thank you, I missed that!

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        int value = 60;
+        Node<int> test = new(value);
+
+        Assert.AreEqual<string>(value.ToString(), test.ToString());
+
+    }
+    /*
+    [TestMethod]
+    public void NodeAppendPass()
```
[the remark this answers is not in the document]
DaisyPalencia: yeah we couldn't get that test to work before Monday, so it was commented out. Thank you, we will go back and fix that!

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,66 @@
...
+        public bool Exists(T value)
+        {
+            var currentNode = this;
+            while (currentNode != null && value != null)
+            {
+                if (currentNode.Equals(value))
+                {
+                    return true;
+                }
+                else if (currentNode.Next != null)
```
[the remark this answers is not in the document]
DaisyPalencia: okay, fixed!

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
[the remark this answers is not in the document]
DaisyPalencia: tests added for exists and appends and clear

DaisyPalencia:
> main things to fix 
-the lack of a Directory.Build.Props file, 
-the 2 test csproj files -the lack of testing on append and exists 
-missing comment about garbage collection

Added the comment for garbage collection, added Directory.Build.Props file, removed second .csproj. 

Thank you for your comments they were very helpful and made our code a lot clearer and cleaner.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+pointer is pointing to something that should no longer
+exist. 
+*/
+namespace GenericsHomework;
+
+public class Node<T>
+{
+    public T Value { get { return valueActual; } }
+
+    public Node<T> next;
```
mmwoodfo, line 23:
```suggestion
    private Node<T> _Next;
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+    }
+    public Node<T> Next
+    {
+        get { return next; }
+        private set { next = value; }
+    }
+    public override string ToString()
+    {
+        return valueActual?.ToString() ?? "is null";
+    }
```
mmwoodfo, line 40:
```suggestion
    public override string? ToString()
    {
        return valueActual?.ToString();
    }
```

PR #128, opened 2023-02-14
Assignment 5 - Ian Kaiserman

Afterlyte:
Paired with @eroadhouse

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -1,0 +1,79 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void Node_CreateStringNode_Success()
+        {
+            Node<string> node = new(value: "test data");
+
+            Assert.IsNotNull(node);
+        }
+
+        [TestMethod]
+        public void ToString_GivenNode_ReturnValue()
+        {
+            Node<int> node = new(value: 1);
+
+            Assert.AreEqual<string>("1", node.ToString()!);
+        }
+
+        [TestMethod]
+        public void Next_GivenNode_ReturnSelf()
+        {
+            Node<string> node = new(value: "test next");
+
+            Assert.AreEqual<Node<string>>(node, node.Next);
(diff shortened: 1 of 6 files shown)


Howard404:
I will review this

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,79 @@
...
+        [TestMethod]
+        public void Next_GivenNode_ReturnSelf()
+        {
+            Node<string> node = new(value: "test next");
+
+            Assert.AreEqual<Node<string>>(node, node.Next);
+        }
+
+        [TestMethod]
+        public void Append_GivenValue_CreatesNewNode()
```
Howard404, line 32: Maybe make a similar test where it appends multiple nodes and returns their values.

Howard404 (commented):
  - Define the `Node` class

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌✔

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌✔

## Fundamentals

No notes on the requirements. Great work!

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
Afterlyte: That's probably a good idea, thank you for the suggestion!

mhtran23:
Define the Node class

Fundamentals
and enabled .NET analyzers for both projects  ✔

Everything looks good.  The code is straight forward and well explained.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,64 @@
+﻿namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public Node(T value)
+        {
+            Value = value;
```
mmwoodfo, line 7:
```suggestion
            Value = value;
            Next = this;
```
It is a good idea to explicitly set next to `this` here in the constructor.

mmwoodfo (commented):
Good job!

GenericsHomework/GenericsHomework/Node.cs
Afterlyte: Interesting, good to know thank you.

PR #129, opened 2023-02-14
Assignment5 (extra credit attempted)

gknutson1:


zlawson-ewu:
I will review this.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,156 @@
...
+    {
+        Append(item);
+    }
+
+
+    // We don't need to worry about a circular loop not being garbage collected
+    // because we set `Next` to `this` for every `Node`, breaking the loop
+    public void Clear()
+    {
+        var node = Next;
```
zlawson-ewu, line 49:
```suggestion
       
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,156 @@
...
+        Next = this;
+        if (node != this) Next.Clear();
```
zlawson-ewu, line 51:
```suggestion
        
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    public void Add(T item)
+    {
+        Append(item);
+    }
+
+
+    // We don't need to worry about a circular loop not being garbage collected
+    // because we set `Next` to `this` for every `Node`, breaking the loop
```
zlawson-ewu, line 46: I don't think you even need to go that far, since 'Next' is the only way to reach the other nodes in the list, the garbage collector won't be able to find a valid root path for them and they'll be collected as a result anyways.

zlawson-ewu (commented):
Instructions:

        Define the Node class ✔

Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following: ✔

Create an Exists method to test to see if a value exists in the list. ❌

Extra Credit:
Implement System.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals:
Place all shared project properties into a Directory.Build.Props file. ❌
and enabled .NET analyzers for both projects ❌ (instructor duplicate)
All of the above should be unit tested (❌?✔)

Commentary:
The list will function but I think some things were overlooked because of limited testing. I've included detail about it in the line comments as well as my altered tests. ICollection is implemented (and has methods that contain the logic needed to fix Exists()), but not fully, according to coverlet. No Directory.Build.Props file, but no X for the '.targets' file because I'm not sure we even needed it for this assignment. I'm not sure if simply including the NetAnalyzers enables them, lang version also not specified. The Asserts.AreEquals should be the generic versions even though the compiler says you don't need to. 

As I've said in the line comments, the list is very close to working, it primarily needs a loop in the Exists() method.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    public override string? ToString()
+    {
+        return Item?.ToString();
+    }
+
+    public bool Exists(T item)
+    {
+        return Equals(item, Item) || (Next != this && Next.Exists(item));
```
eroadhouse, line 121: yeah here you need some kind of loop to iterate through all your nodes otherwise you'll just get stuck when you call this method

eroadhouse (commented):
Overall pretty good, just need to fix the Exists method

Define the Node class
Create an Exists method to test to see if a value exists in the list. ❌
Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ❌

Extra Credit


Implement System.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals
For this assignment, always use Assert.AreEqual<T>() (the generic version) ❌✔

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    public Node(T item)
+    {
+        Item = item;
+    }
+
+    private Node(T item, Node<T> next)
+    {
+        Item = item;
+        Next = next;
```
mmwoodfo, line 18: You should just rely on your append method, what if you only want a single node in your list?
```suggestion
        Next = this;
```

GenericsHomework/GenericsHomeworkTests/GenericsHomeworkICollection.cs
```diff
@@ -0,0 +1,64 @@
+﻿using GenericsHomework;
+using NUnit.Framework;
+
+#pragma warning disable NUnit2005
+namespace GenericsHomeworkTests;
+
+public class GenericsHomeworkICollection
```
mmwoodfo, line 7: I'd recommend adding some tests for ToString as well as some tests for various list lengths.

mmwoodfo (commented):
You had alot of good code review comments left on your code, make sure you check those!

PR #130, opened 2023-02-14
Assignment5

jstanley6:


GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        }
+
+        public bool Exists(T value)
+        {
+            var current = this;
+            do
+            {
+
+                if (current.Value is null && value is null)
```
Joshua-Pedersen, line 45: It doesn't seem that you compare your ``current.Value`` to ``value`` anywhere here. The only time that this method will return true is if you are entering a null value and are comparing an empty node.

GenericsHomework/GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    [TestMethod]
+    public void CreateNodeString()
+    {
+        string testNodeVal = "Hello";
+        Node<string> node = new(testNodeVal);
+        Assert.AreEqual(testNodeVal, node.ToString());
+    }
+
```
Joshua-Pedersen, line 43: Could use a lot more tests, your Append, Clear, and Exist methods are not being tested.

Joshua-Pedersen (commented):
## General

- Lots of files and folders that don't need to be pushed, seems you committed your entire fork.

## Instructions

  - Define the `Node` class
  - Create an Exists method to test to see if a value exists in the list. ❌

## Extra Credit

## Fundamentals

- All of the above should be unit tested ❌

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,61 @@
...
+            Node<T> newNode = new Node<T>(value);
+            newNode.NextNode = NextNode;
+            NextNode = newNode;
+
+            return newNode;
+
+        }
+
+        public bool Exists(T value)
+        {
```
mmoran314, line 40: Would redo this Exists method, mostly checking the if statement about the "value is null"

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,61 @@
+﻿namespace GenericsHomework;
```
mmoran314, line 1: Need to add comment about the need to use garbage collector

mmoran314 (commented):
Define the Node class
Place all shared project properties into a Directory.Build.Props file. ✔

Looks good!
First thing I would do is add the comment in regard to needing to use a garbage collector for the required bullet point.
Even though there is an Exists method, it seems like it could use some work.

mmwoodfo (commented):
You have alot of changed files, try to keep your git difference clean and only have the files related to the assignment requirements show up.

PR #134, opened 2023-02-20
Assignment6 lambda expressions - Hunter Thompson & Hunter Allen

StrangeRanger:
@hallen10


zlawson-ewu:
I will review this.

Calculate/ProgramMain.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Calculate;
+
```
zlawson-ewu, line 2: To be consistent with the naming specs, This should probably just be "Program.cs", and the other should be called "ProgramBase.cs".

Calculate/Calculator.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (! MathematicalOperations.ContainsKey(operation)) { return false; }
+        } 
+        catch (Exception e) when (e is IndexOutOfRangeException or FormatException)
+        {
+            return false;
+        }
+
+        return true;
+    }
+
```
zlawson-ewu, line 56: There's a lot of logic here also repeated in Calculate(). You could simplify Calculate() by using a call to TryCalculate() inside of it, but really you can move the calculating logic and extra parameter to TryCalculate() instead and it should still work. This scheme would've likely been easier if you had passed in a string instead, and returned a bool along with an out variable for the math result, moving the user input stuff to Main in ProgramMain.cs. All the same, this calculator does function as intended when I run it.

Calculate/ProgramMain.cs
```diff
@@ -0,0 +1,13 @@
...
+public class ProgramMain
+{
+    public static void Main()
+    {
+        Program program = new();
+        Calculator calculator = new();
+        
+        calculator.Calculate(program.WriteLine, program.ReadLine);
+
```
zlawson-ewu, line 11: Though not required, the lack of a prompt threw me when I went to test it, since you're only faced with a blank console window . You might also consider letting the user use the program until an 'exit' signal is sent. Again though, neither of these were requirements.

Calculate.Tests/ProgramTest.cs
```diff
@@ -0,0 +1,42 @@
+namespace Calculate.Tests;
+
```
zlawson-ewu, line 2: Just a heads-up, I have 5 warnings across ProgramTest.cs and CalculatorTest.cs related to null, I think most of it can be solved with "?" or "!", but you'll want to get rid of those warnings before this gets graded.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,83 @@
...
+        return (double) a / b;
+    }
+
+    // TODO: Should probably be re-written...
+    public bool TryCalculate(Func<string> readLine)
+    {
+        // TODO: ???
+        string input = readLine() ?? throw new ArgumentNullException(typeof(string).ToString());
+        string[] parts = input.Split(' ');
+
```
zlawson-ewu, line 40: I'd suggest making sure there are -only- three objects in the array after the Split(), any less and it's an invalid calculation and you'll get Index exceptions (which you have accounted for, to be fair), any more and you might parse a string that starts with a valid operation but can continue beyond that.

zlawson-ewu (commented):
# Assignment

The purpose of this assignment is to learn how to work with delegates and lambda expressions. We will define a 
calculator class that stores 4 actions corresponding to +, -, *, /, and multiple. The calculator class will
also support customizable behavior for reading and writing content.

## Instructions

- Create a *Console* project called "Calculate.". ✔
- Define a Program Class
  - Define two init-only setter properties, `WriteLine` and `ReadLine`, that contain delegates for writing a line of text and reading a line of text respectively ✔
  - Write a test that sets these properties at construction time and then invokes the properties and verifies the expected behavior occurs. ❌
  - Set the default behavior for the `WriteLine` and `ReadLine` properties to invoke `System.Console` versions of the methods and add an empty default constructor. (❌?✔)
- Define a Calculator class ✔
  - Define static `Add`, `Subtract`, `Multiple`, and `Divide` methods that have two parameters and return a third parameter. ✔
  - Define a read-only property, `MathematicalOperations`, of type `System.Collections.Generics.IReadOnlyDictionary<TKey,TValue>` that:
    - is initialized to a `System.Collections.Generics.Dictionary<<TKey,TValue>` instance that. ✔
      - Uses `char` for the key corresponding to the operators +, -, *, and /. ✔
      - Has values that correspond with the Add, Subtract, Multiple, and Divide methods. ✔
  - Implement a `TryCalculate` method following "TryParse" pattern ✔
    - Valid `calculation` expressions include such strings as "3 + 4", "42 - 2", etc. ✔
    - If there is no whitespace around the operator, you can assume the `calculation` is invalid and return false. Similarly if the operands are not integers. ✔
    - Use `string.Split()`, pattern matching, logical and operators to parse the string in their entirety ✔
    - Index into the `MathematicalOperations` method using the operator parsed during pattern matching to find the corresponding implementation and invoke it. ✔
- Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ✔
- Be sure to use the `WriteLine`/`ReadLine` properties on `Program` for testing the input and output of your program. ✔

## Extra Credit

- Refactor the redirect portion of the `Program` class into 'ProgramBase` (❌?✔)
- Move ProgramBase into a ConsoleUtilities assembly to be used in other console-based projects
- Use generics the mathematical operations methods and consider using generic constraints (requires .NET 7.0)

## Fundamentals

## Commentary

- Overall, good work, the calculator does function, and the code is simple for the most part. It looks like there was some confusion when the TryCalculate() method was written, so you might want to clean that up. I've left line comments everywhere I thought relevant. As said in one of them, I did get 5 warnings when building the project, so those should be attended to if you don't want points off.

Calculate/Program.cs
```diff
@@ -0,0 +1,7 @@
+namespace Calculate;
+
+public class Program
+{
+    public Action<string> WriteLine { get; init; } = s => Console.WriteLine(s);
+    public Func<string?> ReadLine { get; init; } = () => Console.ReadLine();
```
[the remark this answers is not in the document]
StrangeRanger: Is that really necessary, though? The default constructor always gets created during compile time, even when it's not explicitly specified.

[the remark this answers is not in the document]
zlawson-ewu: In this case, I marked this because the bullet point explicitly asked for one to be added, similarly to the Assert.AreEqual<>() generic thing. It might not be necessary in reality, but I'm just going down the list when I'm doing these reviews, and off of my own interpretation of the requirements as well. It might be something worth asking about in class or on Teams.

Calculate/Calculator.cs
StrangeRanger: I did that initially but it resulted in a problem where the program asked for input twice because I had to account for if someone called TryCaculate directly or through the Calculate method. I'll see if I can make it work but use a different strategy.

Calculate/Program.cs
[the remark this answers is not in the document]
StrangeRanger: Alright, that makes sense. Thanks for pointing it out.

Calculate.Tests/ProgramTest.cs
```diff
@@ -0,0 +1,42 @@
...
+    [TestMethod]
+    public void Program_ReadLine()
+    {
+        StringReader stringReader = new("Hello World");
+        Console.SetIn(stringReader);
+        
+        string? actual = Program.ReadLine();
+
+        Assert.AreEqual("Hello World", actual);
+    }
```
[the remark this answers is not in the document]
StrangeRanger: I see. I went ahead and did something similar, but separated it out as I have it in the two other Program tests.

Calculate/ProgramMain.cs
StrangeRanger: I'll at least at a prompt telling the user to input some calculation

Calculate/Calculator.cs
StrangeRanger: Was able to figure out a decent way of doing it

Calculate/ProgramBase.cs
```diff
@@ -0,0 +1,13 @@
+namespace Calculate;
+
+public class ProgramBase
+{
+    public Action<string> WriteLine { get; init; } = s => Console.WriteLine(s);
```
mmwoodfo, line 5:
```suggestion
    public Action<string> WriteLine { get; init; } = Console.WriteLine;
```
This should work the same way.

Calculate/ProgramBase.cs
```diff
@@ -0,0 +1,13 @@
...
+    public Func<string?> ReadLine { get; init; } = () => Console.ReadLine();
```
mmwoodfo, line 6: Same here
```suggestion
    public Func<string?> ReadLine { get; init; } = Console.ReadLine;
```

Calculate/ProgramBase.cs
```diff
@@ -0,0 +1,13 @@
...
+
+    // ASSIGNMENT QUESTION: Is this empty constructor needed? The default constructor is created by
+    //                      default during compilation, right? Wouldn't that make this constructor
+    //                      redundant?
+    public ProgramBase()
```
mmwoodfo, line 11: This is not required if it is your only constructor, however, the assignment asked that you also define a constructor for setting the Write and ReadLine properties. So, if you had a defined constructor such as that, you would need to explicitly define a default constructor if you wanted to allow people to still instantiate this class with default behavior.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,72 @@
...
+        { '+', Add },
+        { '-', Subtract },
+        { '*', Multiply },
+        { '/', Divide }
+    };
+
+    public static double Add(int a, int b)
+    {
+        return a + b;
+    }
```
mmwoodfo, line 18:
```suggestion
    public static double Add(int a, int b) => a + b;
```
You could simplify this into a single-lined lambda. Feel free to write methods on a single line if they are just performing a single operation such as this.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return a * b;
+    }
+
+    public static double Divide(int a, int b)
+    {
+        return (double)a / b;
+    }
+    
+    public bool TryCalculate(string input)
```
mmwoodfo, line 35: You'll need an out parameter here. e.g.
```suggestion
    public bool TryCalculate(string input, out result)
```

Calculate/Calculator.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        string[] parts = input.Split(' ');
+
+        try
+        {
+            _ = int.Parse(parts[0]);
+            _ = int.Parse(parts[2]);
+            char operation = parts[1].ToCharArray()[0];
```
mmwoodfo, line 43: You'll want to actually return the result of the operation here as well by setting your `out` parameter, `result`, equal to the operation's outcome.

eg:
`result = MathamaticalOperations[operation](a, b);`

Calculate/Calculator.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            if (!MathematicalOperations.ContainsKey(operation)) { return false; }
+        }
+        catch (Exception e) when (e is IndexOutOfRangeException or FormatException)
+        {
+            Console.Error.WriteLine(e.Message);
```
mmwoodfo, line 49: I wouldn't use a console log here

Calculate/Calculator.cs
```diff
@@ -0,0 +1,72 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    public void Calculate(Action<string> writeLine, Func<string?> readLine)
```
mmwoodfo, line 56: I think the above method is supposed to do the calculations, you shouldn't need this method at all.

Calculate.Tests/CalculatorTest.cs
```diff
@@ -0,0 +1,122 @@
...
+    public void TestInitialize()
+    {
+        TestCleanup();
+        Calculator = new Calculator();
+    }
+
+    [TestCleanup]
+    public void TestCleanup()
+    {
+        Calculator = null !;
```
mmwoodfo, line 20:
```suggestion
        Calculator = null!;
```

Calculate.Tests/CalculatorTest.cs
```diff
@@ -0,0 +1,122 @@
...
+public class CalculatorTest
+{
+    #pragma warning disable CS8618
+    private Calculator Calculator { get; set; }
+    #pragma warning restore CS8618
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        TestCleanup();
```
mmwoodfo, line 13: Your test clean up should be getting called at  the end of each test.

PR #136, opened 2023-02-20
Assignment6 @lsobtafo

slanante:
Initial PR, will add more tests.
Had issues with sending a PR since the set up of the repository was incorrect.
Here are the commits from the original branch:
image

### DIFF
--- a/Calculate.Tests/Calculate.Tests.csproj
+++ b/Calculate.Tests/Calculate.Tests.csproj
@@ -1,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageReference Include="MSTest.TestAdapter" Version="2.2.10" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.2.10" />
+    <PackageReference Include="coverlet.collector" Version="3.1.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Calculate\Calculate.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/Calculate/Calculator.cs
+++ b/Calculate/Calculator.cs
@@ -1,0 +1,52 @@
+namespace Calculate
+{
+    public class Calculator
+    {
+        public static Func<int, int, int> Add { get; set; } = (v1, v2) => v1 + v2;
+        public static Func<int, int, int> Subtract { get; set; } = (v1, v2) => v1 - v2;
+        public static Func<int, int, int> Multiply { get; set; } = (v1, v2) => v1 * v2;
+        public static Func<int, int, int> Divide { get; set; } = (v1, v2) => v1 / v2;
+
+        public static IReadOnlyDictionary<char, Func<int, int, int>> MathematicalOperations { get; } =
+            new Dictionary<char, Func<int, int, int>>()
+            {
+                {'+', Add},
+                {'-', Subtract},
+                {'*', Multiply},
+                {'/', Divide}
+            };
+
+        public static bool TryCalculate(string input, out int result)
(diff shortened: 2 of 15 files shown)


Calculate/Calculator.cs
```diff
@@ -0,0 +1,52 @@
+namespace Calculate
+{
+    public class Calculator
+    {
+        public static Func<int, int, int> Add { get; set; } = (v1, v2) => v1 + v2;
+        public static Func<int, int, int> Subtract { get; set; } = (v1, v2) => v1 - v2;
+        public static Func<int, int, int> Multiply { get; set; } = (v1, v2) => v1 * v2;
+        public static Func<int, int, int> Divide { get; set; } = (v1, v2) => v1 / v2;
```
jacobnewcomb, line 8: I would maybe check if v2 is zero here. If so, throw an error?

jacobnewcomb (commented):
- Create a Console project called "Calculate.". ✔
- Define a Program Class
        Write a test that sets these properties at construction time and then invokes the properties and verifies the expected behavior occurs. ✔
- Define a Calculator class ✔
        Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. ✔
- Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ✔
- Be sure to use the WriteLine/ReadLine properties on Program for testing the input and output of your program. ✔

Calculate/Calculator.cs
jacobnewcomb: for the division specifically

slanante: Cool, will do thank you!

mmwoodfo (commented):
image
Why can't you handle fractions?

image
This input is valid, and would have had a similar output to the one above (decimal)

Calculate/Calculator.cs
```diff
@@ -0,0 +1,84 @@
...
+        public static Func<int, int, int> Multiply { get; set; } = (v1, v2) => v1 * v2;
+        public static Func<int, int, int> Divide { get; set; } = (v1, v2) =>
+        {
+            if (v2 == 0)
+            {
+                throw new DivideByZeroException("Cannot divide by zero");
+            }
+            if (v1 < v2)
+            {
+                throw new InvalidOperationException("Numerator must be greater than or equal to the denominator, we cannot handle fractions");
```
mmwoodfo, line 16: Why? This should be handled.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                return false;
+            }
+
+            if (!Calculator.MathematicalOperations.TryGetValue(symbol, out var operation))
+            {
+                return false;
+            }
+
+            if (char.TryParse(parts[1], out symbol).Equals('/') && (parts[2] == "0"))
```
mmwoodfo, line 59: Isn't this handled by your catch below?

Calculate/ProgramBase.cs
```diff
@@ -0,0 +1,8 @@
+namespace Calculate
+{
+    public abstract class ProgramBase
```
mmwoodfo, line 3: This should have a constructor to set these properties, as well as a default constructor to allow for the default implementation.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,84 @@
+namespace Calculate
+{
+    public class Calculator
+    {
+        public static Func<int, int, int> Add { get; set; } = (v1, v2) => v1 + v2;
```
mmwoodfo, line 5:
```suggestion
        public static readonly Func<int, int, int> Add { get; set; } = (v1, v2) => v1 + v2;
```

Calculate.Tests/Calculate.Tests.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
```
mmwoodfo, line 4: Should be taken care of in your directory build props file

Calculate/Calculator.cs
```diff
@@ -0,0 +1,84 @@
...
+                {'-', Subtract},
+                {'*', Multiply},
+                {'/', Divide}
+            };
+
+        public static bool TryCalculate(string input, out int result)
+        {
+            result = 0;
+
+            string[] parts = input.Split(' ');
```
mmwoodfo, line 35: You should be using the Split method along side pattern matching. e.g.

PR #137, opened 2023-02-20
Assignment 6 - Zach Lawson and Tom Rohr

zlawson-ewu:
Assignment 6 peer programmed by Zach Lawson (@zlawson-ewu) and Tom Rohr (@Tom-Rohr). First two extra-credit bullet points completed.


Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Calculate;
+
+public class Calculator
+{
+    public bool TryCalculate(in string calculation, out double result)
+    {
+        bool isValidCalculation = false;
```
lsobtafo, line 7: The isValidCalculation variable can be simplified by initializing it to 'true' and only setting it to 'false' if any of the conditions for a valid calculation fail.

Calculate/Calculate/Program.cs
```diff
@@ -0,0 +1,26 @@
+﻿using ConsoleUtilities;
+namespace Calculate;
+
+public class Program : ProgramBase
+{
+    public static void Main()
+    {
+        Program prog = new();
```
Tom-Rohr, line 8: After doing my peer review I think we need to change this up a little to meet the following criteria. Basically just declare the calc outside of main.

_Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ❌✔_

Calculate/Calculate/Calculator.cs
zlawson-ewu: I think that would result in the same number of lines being used. I have it default to 'false' to follow what I consider 'positive validation', which is to say that something should be considered invalid unless it can be proven valid, whereas start from 'true' instead would follow 'negative validation'. The general principle is to avoid false-positives, and having 'true' as the default value might let one through, but starting from false never will.

Calculate/Calculate/Program.cs
zlawson-ewu: Main() is just a method within Program.cs, so that requirement _should_ be met, though it might be better practice to initialize it as a static readonly member of the class, since nothing inside it ever needs to change.

Calculate/ConsoleUtilities/ProgramBase.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace ConsoleUtilities;
+
+public class ProgramBase
+{
+    public Action<string> WriteLine { get; init; } = Console.WriteLine;
+    public Func<string?> ReadLine { get; init; } = Console.ReadLine;
+
+    public ProgramBase() { }
```
mmwoodfo, line 8: There should also be a constructor here for setting the write/read properties.

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,38 @@
...
+        if (tokens.Length == 3)
+        {
+            if (int.TryParse(tokens[0], out int x)
+                && tokens[1].Any(op => MathematicalOperations.Keys.Contains(op))
+                && char.TryParse(tokens[1], out char opChar)
+                && int.TryParse(tokens[2], out int y))
+            {
+                isValidCalculation = true;
+                Func<int, int, double> mathOperation = MathematicalOperations[opChar];
+                result = mathOperation(x, y);
```
mmwoodfo, line 19: also acceptable to just return true after this line, and remove the `isValidCalculation` all together

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Calculate;
+
+public class Calculator
+{
+    public bool TryCalculate(in string calculation, out double result)
+    {
+        bool isValidCalculation = false;
+        result = default;
+        string[] tokens = calculation.Split(' ');
```
mmwoodfo, line 9: You should be using the Split method along side pattern matching. e.g.

mmwoodfo (commented):
Good job!

Calculate/Calculate/Calculator.cs
zlawson-ewu: Good to know, I think I used the bool here for readability and because one of the guidelines says to avoid multiple returns. With 'isValidCalculation' heading each section (the start of the method and the 'valid' block), it functions somewhat like a title for the sections.

Calculate/CalculateTests/CalculatorTests.cs
```diff
@@ -0,0 +1,89 @@
+﻿using Calculate;
+
+namespace CalculateTests;
+
+[TestClass]
+
+public class CalculatorTests
```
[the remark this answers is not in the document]
zlawson-ewu: Yeah, looking back at it we ought to have included a few more tests for valid input. The divide-by-0 one I didn't consider, as I tested that manually early on and discovered that it just returns the infinity symbol. Since it had a valid output already, I thought of the case as being the same for any other valid input, but for completeness I agree that it should have been checked.

Calculate/ConsoleUtilities/ProgramBase.cs
zlawson-ewu: I think we got so hung up on following the specs exactly that this was forgotten. Duly noted.

PR #142, opened 2023-02-21
Assignment6 lambda expressions - Nolan Posey & Josh Merrell

Nolan-Posey:
@Nolan-Posey
@joshmerrell

### DIFF
--- a/Calculate/Calculate/Calculator.cs
+++ b/Calculate/Calculate/Calculator.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Calculate
+{
+    public class Calculator
+    {
+    }
+}
--- a/Calculate/Calculate/Calculator.cs
+++ b/Calculate/Calculate/Calculator.cs
@@ -43,7 +43,7 @@ namespace Calculate
                 {
                     LeftOperand = float.Parse(leftOperand);
                     RightOperand = float.Parse(rightOperand);
-                } catch(FormatException ex) 
+                } catch (FormatException)
                 {
                     result = 0;
                     return false;
--- a/Calculate/Calculate/Calculator.cs
+++ b/Calculate/Calculate/Calculator.cs
@@ -8,5 +8,26 @@ namespace Calculate
 {
     public class Calculator
     {
+        public Calculator() 
+        {
+            MathematicalOperations = new Dictionary<char, Func<float, float, float>>();
+        }
+        public IReadOnlyDictionary<char,Func<float,float,float>>? MathematicalOperations { get; }
+        public static float Add(float leftOperand, float rightOperand) 
+        {
+            return leftOperand + rightOperand;
+        }
+        public static float Subtract(float leftOperand, float rightOperand)
+        {
+            return leftOperand - rightOperand;
+        }
+        public static float Multiply(float leftOperand, float rightOperand)
+        {
+            return leftOperand * rightOperand;
+        }
+        public static float Divide(float leftOperand, float rightOperand)
+        {
+            return leftOperand / rightOperand;
+        }
     }
 }
--- a/Calculate/Calculate/Calculator.cs
+++ b/Calculate/Calculate/Calculator.cs
@@ -29,5 +29,34 @@ namespace Calculate
         {
             return leftOperand / rightOperand;
         }
+        public bool TryCalculate(string equation) 
+        {
+            if (equation.Split(" ") is [string { Length: > 0 } leftOperand, 
+                string { Length: > 0} operatr, string { Length: > 0} rightOperand]) 
+            {
+                float LeftOperand = float.Parse(leftOperand);
+                float RightOperand = float.Parse(rightOperand);
+                switch (operatr) 
+                {
+                    case("+"):
+                        //result = Add(LeftOperand, RightOperand);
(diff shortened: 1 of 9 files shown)


Calculate/Calculate/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class Program
+    {
+        public static void Main(string[] args) 
+        {
+            Program prog = new();
+            Calculator calc = new();
+            float result;
+            bool calculatedSuccessfully;
+
+            prog.WriteLine("Please enter an equation to be solved.");
```
randypipkins, line 18: Could potentially put this in a do;while; loop to continuously ask for user input after each successful calculation. Just a suggestion as the code runs and compiles fine without it but it would keep the user from having to run the code each time they wanted to do a new calculation.

randypipkins:
Code looks good! Good idea using floats instead of ints or doubles as it will provide a more precise answer on the calculations

Calculate/Calculate/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+                prog.WriteLine($"The answer to the equation is: {result}");
+            }
+            else 
+            {
+                prog.WriteLine!("The equation inputed was not in the right format.");
+            }
+        }
+        public Action<string> WriteLine { get; init; } = text => Console.WriteLine(text);
+        public Func<string>? ReadLine { get; init; } = () => Console.ReadLine()!;
+        public Program()
```
mmoran314, line 32: I believe this public Program() section can be deleted.

mmoran314 (commented):
Create a Console project called "Calculate.". ✔
Define a Program Class
Write a test that sets these properties at construction time and then invokes the properties and verifies the expected behavior occurs. ✔
Define a Calculator class ✔
Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. ✔
Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ✔
Be sure to use the WriteLine/ReadLine properties on Program for testing the input and output of your program. ✔

Place all shared project items into a Directory.Build.targets file. ❌✔

It looks really good!
Some small parts may be able to be cleaned up but I think everything is well done and easy to read.

Calculate/Calculate/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Calculate
+{
+    public class Program
+    {
+        public static void Main(string[] args) 
+        {
+            Program prog = new();
```
mmwoodfo, line 13:
```suggestion
            Program program = new();
```
Avoid abbreviations and acronyms in variable names

Calculate/Calculate/Program.cs
mmwoodfo, line 32: There should be a constructor to set the read/write line properties

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,76 @@
...
+            ((Dictionary<char, Func<float, float, float>>)MathematicalOperations).Add('+', Add);
+            ((Dictionary<char, Func<float, float, float>>)MathematicalOperations).Add('-', Subtract);
+            ((Dictionary<char, Func<float, float, float>>)MathematicalOperations).Add('*', Multiply);
+            ((Dictionary<char, Func<float, float, float>>)MathematicalOperations).Add('/', Divide);
+        }
+        public IReadOnlyDictionary<char, Func<float, float, float>>? MathematicalOperations;
+        public static float Add(float leftOperand, float rightOperand) 
+        {
+            return leftOperand + rightOperand;
+        }
```
mmwoodfo, line 24:
```suggestion
        public static float Add(float leftOperand, float rightOperand) => leftOperand + rightOperand;
```
You can shorten thses!

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            return leftOperand * rightOperand;
+        }
+        public static float Divide(float leftOperand, float rightOperand)
+        {
+            return leftOperand / rightOperand;
+        }
+        public bool TryCalculate(string equation, out float result) 
+        {
+            if (equation.Split(" ") is [string { Length: > 0 } leftOperand,
```
mmwoodfo, line 39: You should be using the Split method along side pattern matching. e.g.

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,76 @@
...
+                try 
+                {
+                    LeftOperand = float.Parse(leftOperand);
+                    RightOperand = float.Parse(rightOperand);
+                } catch (FormatException)
+                {
+                    result = 0;
+                    return false;
+                }
+                switch (operatr)
```
mmwoodfo, line 53: Instead of doing a switch here, you should be looking up the operation via the dictionary key, which is the character operator (+, -, *, /)

Calculate/CalculateTests/CalculatorTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        [TestMethod]
+        public void Calculator_TryCalculateGivenBadInput_ReturnsFalse()
+        {
+            //Arrange
+            Calculator calc = new();
+            float result;
+
+            //Act
+            string equation1 = "42 +! 3";
```
mmwoodfo, line 38: This is a great spot for data driven tests!

PR #147, opened 2023-02-21
Assignment 6: @crawfordkid2 and @GeorgeKoralski007

GeorgeKoralski007:
First implementation


slanante (commented):
Overall, good job. There are some missing portions such as the Tests for the WriteLine/ReadLine properties of the Program class. I see the current tests utilizing these properties by giving the Calculator a string such as "4 - 2" but I think it would be good to test outside of a Calculator context. For example, test that "I am a random string" is both written and read. I also noted work that could be done on the Calculator class as well.

Create a Console project called "Calculate.". ✔
Define a Program Class
Write a test that sets these properties at construction time and then invokes the properties and verifies the expected behavior occurs. ❌
Define a Calculator class ✔
Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. ❌
Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ✔
Be sure to use the WriteLine/ReadLine properties on Program for testing the input and output of your program. ✔

Calculate/Calculate/Program.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Calculate;
+public class Program
+{
+    public Action<string> WriteLine { get; init; } = x => Console.WriteLine(x);
+    public Func<string?> ReadLine { get; init; } = () => Console.ReadLine();
+
+    public Program() { }
```
mmwoodfo, line 7: You should have a constructor for setting the write/read properties as well as the default constructor.

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,40 @@
...
+            ['/']= Divide,
+        };
+
+        public bool TryCalculate(string input, out int output)
+        {
+            output = 0;
+            if (string.IsNullOrEmpty(input))
+                return false;
+
+            string[] calc = input.Split(' ');
```
mmwoodfo, line 26: You should be using the Split method along side pattern matching. e.g.

PR #148, opened 2023-02-21
Assignment6 alan and grant

arose-13:
Alan Rose (@arose-13) and Grant Howard (@Howard404)


BenFristadSchool (commented):
I will review this

Calculate/Calculate/Calculate.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
BenFristadSchool, line 4: Remove duplicate XML handled in Directory.Build.props

Calculate/Calculate.Tests/Calculate.Tests.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
BenFristadSchool, line 6: Remove duplicate XML handled in Directory.Build.props

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Calculate;
+
+public class Calculator
+{
+    public Action<string> WriteLine { get; init; }
```
BenFristadSchool, line 11: The ReadLine and WriteLine properties should be defined in Program.cs

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,89 @@
...
+        = new Dictionary<char, Func<double, double, double>>
+        {
+            { '+', Add },
+            { '-', Subtract },
+            { '*', Multiple },
+            { '/', Divide },
+        };
+
+    public Calculator()
+    {
```
BenFristadSchool, line 23: Calculator and Program classes should be independent of each other

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,89 @@
...
+        ReadLine = System.Console.ReadLine;
+    }
+
+    public Calculator(Action<string> writeLine, Func<string?> readLine)
+    {
+        WriteLine = writeLine;
+        ReadLine = readLine;
+    }
+
+    public static double Add(double a, double b) => a + b;
```
BenFristadSchool, line 34: The operands are supposed to be integers

Calculate/Calculate.Tests/CalculatorTests.cs
```diff
@@ -0,0 +1,181 @@
...
+using static System.Net.Mime.MediaTypeNames;
+
+namespace Calculate.Tests;
+
+[TestClass]
+public class CalculatorTests
+{
+    [TestMethod]
+    public void Create_UsingDefaultConstructor_Success()
+    {
```
BenFristadSchool, line 12: ReadLine and WriteLine tests should probably be written in a Program.Tests.cs file

Calculate/Calculate.Tests/CalculatorTests.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        calculator.Calculate("3 * 3");
+        Assert.AreEqual<double>(result, 9);
+
+        calculator.Calculate("3 / 3");
+        Assert.AreEqual<double>(result, 1);
+    }
+
+    [TestMethod]
+    public void Calculate_GivenOperationWithRemainder_WritesDecimal()
```
BenFristadSchool, line 75: Its supposed to only work with integers.
```If there is no whitespace around the operator, you can assume the calculation is invalid and return false. Similarly if the operands are not integers. ❌✔```

File: Calculate/Calculate.Tests/CalculatorTests.cs [hunk 8]
```diff
@@ -0,0 +1,181 @@
...
+        calculator.WriteLine("Hello World!");
+
+        // Assert
+        Assert.AreEqual<string>(text, "Hello World!");
+        
+    }
+
+    [TestMethod]
+    public void Calculate_GivenValidOperations_AllOperationsReturnCorrectResult()
+    {
```
BenFristadSchool, line 51: You should write each operation as a different test so the tests are more descriptive if they were to fail.
For example, there should be a test for add and a separate test for subtract. You can also set up tests for when they will fail.

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            if (operate == '/' && b == 0)
+                throw new DivideByZeroException();
+            else if (!operand1 || !operand2 || !operation)
+                throw new FormatException();
+            else
+                WriteLine($"{MathematicalOperations[operate](a, b)}");
+
+        }
+        catch (FormatException)
```
BenFristadSchool, line 68: All these extra cases do bring some value, but it does add complexity, which is discouraged in the assignment instructions. Its not wrong, but I would choose wisely where you add extra complexity.

BenFristadSchool (commented):
Well Done!

arose-13:
> Well Done!

Thank you for your feedback! We will get those things fixed :)

Calculate/Calculate/Calculator.cs
arose-13: I will try to simplify. Thanks!

Calculate/Calculate.Tests/CalculatorTests.cs
arose-13: That would be a better way of handling this -- creating separate tests for each operation would be better.

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            if (operation == '/' && rightOperand == 0)
+                throw new DivideByZeroException();
+            else if (!containsLeft || !containsRight || !containsOperation || calcuationParts.Length > 3)
+                throw new FormatException();
+            else
+                return $"{MathematicalOperations[operation](leftOperand, rightOperand)}";
+
+        }
+        catch (FormatException)
```
mmwoodfo, line 40: You can catch multiple exceptions,
```suggestion
        catch (Exception ex) when (ex is FormatException || ex is DivideByZeroException || ex is KeyNotFoundException)
        {
           return ex.message;
        }

```

Calculate/Calculate/Calculator.cs
```diff
@@ -0,0 +1,57 @@
...
+    public static int Divide(int leftOperand, int rightOperand) => leftOperand / rightOperand;
+
+    public string TryCalculate(string? calculation)
+    {
+        if (string.IsNullOrWhiteSpace(calculation))
+            return "Error: Cannot accept a null operation!";
+
+        try
+        {
+            string[] calcuationParts = calculation.Split(' ');
```
mmwoodfo, line 26: You should be using the Split method along side pattern matching. e.g.

Calculate/Calculate.Tests/CalculatorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+        {
+            WriteLine = (string input) => { result = int.Parse(input); },
+        };
+
+        // Act and Assert
+        program.WriteLine(program.calculator.TryCalculate("3 + 3"));
+        Assert.AreEqual<int>(result, 6);
+
+        program.WriteLine(program.calculator.TryCalculate("823 + 100"));
+        Assert.AreEqual<int>(result, 923);
```
mmwoodfo, line 22: I think you got some good feedback, that the read and writeline methods should be in the program test. Here you should focus on methods only inside the calculator class.

Calculate/Calculate.Tests/CalculatorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        // Act and Assert
+        program.WriteLine(program.calculator.TryCalculate("3 * 3"));
+        Assert.AreEqual<int>(result, 9);
+
+        program.WriteLine(program.calculator.TryCalculate("350 * 100"));
+        Assert.AreEqual<int>(result, 35000);
+
+        program.WriteLine(program.calculator.TryCalculate("0 * 0"));
+        Assert.AreEqual<int>(result, 0);
```
mmwoodfo, line 73: This is a great place to use data driven tests

PR #156, opened 2023-02-28
Assignment6 @mhtran23

BenagainB:


Calculate/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+
+
+namespace Calculate
+{
+    public class Program
+    {
+        public Action<string> WriteLine { get; init; } = Console.WriteLine;
+        public Func<string?> ReadLine { get; init; } = Console.ReadLine;
+
+        public Program() { }
```
mmwoodfo, line 15: You should have a constructor that allows you to set the read/write properties too

Calculate/Calculator.cs
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+        public static bool TryCalculate(string calculation, out int result)
+        {
+            result = 0;
+            int operand1;
+            int operand2;
+            char operator1;
+
+            string[] input = calculation.Split(' ');
```
mmwoodfo, line 39: You should be using the Split method along side pattern matching. e.g.

Calculate/Directory.Build.Props
```diff
@@ -0,0 +1,9 @@
+﻿<Project>
+  <PropertyGroup>
```
mmwoodfo, line 2: This should be in your solution items folder

Calculate.Tests/CalculatorTest.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Calculate.Tests
+{
+    [TestClass]
+    public class CalculateTests
```
mmwoodfo, line 6: This is a good case for test driven tests, so you can test multiple number combinations.

mmwoodfo (commented):
If you don't get points back for your PR reviews on assignment 7&8, just shoot me an email.

PR #167, opened 2023-03-07
Assignment7+8 Clockknight Malcolm-Shepherd

Clockknight:


BenFristadSchool (commented):
I will review this

Assignment/Assignment/Assignment.csproj
```diff
@@ -4,4 +4,10 @@
     <ProjectReference Include="..\Assignment.Interfaces\Assignment.Interfaces.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="People.csv">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
BenFristadSchool, line 11: Where is ```<PropertyGroup>``` and Directory.Build.props is also missing

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,95 @@
...
+using System.Linq;
 
 namespace Assignment
 {
     public class SampleData : ISampleData
     {
         // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        public IEnumerable<string> CsvRows 
+        {
```
BenFristadSchool, line 13: This can be written more concisely. You should be able to get away with removing the ```where line.Contains('@')``` line and just read all the lines and skip line 1

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,245 @@
...
+            foreach (string state in statesArr)
+            {
+                Assert.IsTrue(!(unique.Contains(state)));
+                unique.Add(state);
+            }
+        }
+    }
+
+    [TestClass]
+    public class NodeEnumerableTests
```
BenFristadSchool, line 224: Its just a suggestion, but you could move over your Node.tests from your old project and add these node tests to it for better organization.

Assignment/Assignment/Node.cs
```diff
@@ -0,0 +1,111 @@
...
+            // At least once all local instances of other nodes pass out of scope.
+        }
+
+
+        public IEnumerable<Node<T>> ChildItems(int maximum)
+        {
+            List<Node<T>> child = new();
+            Node<T> currentNode = _last;
+
+            int counter = 1;
```
BenFristadSchool, line 83: Another alternative to this method that will allow you to initialize less variables is to decrement the ```int maximum``` variable rather than increment ```int counter```. Just a slight change

Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+    [TestClass]
+    public class GetUniqueSortedListOfStatesGivenCsvRowsTests
+    {
+        /*    [TestInitialize]
+            public void AssemblyInitialize()
+            {
+                //source = csvrows()
+            }*/
```
BenFristadSchool, line 43: You should either implement this or remove the comment

BenFristadSchool (commented):
Well Done!

Assignment/Assignment/Assignment.csproj
Malcolm-Shepherd: There is a Directory.Build.Props under the Solution Items tab

Assignment/Assignment/SampleData.cs
Malcolm-Shepherd: I didn't realize there was a skip method, that makes a lot more sense.

Assignment/Assignment.Tests/SampleDataTests.cs
Malcolm-Shepherd: Deleted all these commented out blocks

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
+            } 
+        }
 
         // 2.
-        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+            List<string> states = new List<string>();
+            foreach(string item in CsvRows)
```
mmwoodfo, line 27: Avoid using foreach here, you can use select instead!

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,94 @@
...
 
         // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+        public IEnumerable<IPerson> People
+        { 
+            get 
+            {
+                List<Person> people = new List<Person>();
+                List<Address> addresses = new List<Address>();
+                foreach (string item in CsvRows)
```
mmwoodfo, line 52: Same here!

mmwoodfo (commented):
I don't see a test with a hardcoded list of addresses. I also don't see any tests for your node class.

PR #168, opened 2023-03-07
Assignment7+8 - Andrew Korchemniy & Vera Korchemnaya

AndrewKorchemniy:
@VeraKorchemnaya

### DIFF
--- a/Assignment/Assignment.Tests/SampleDataTests.cs
+++ b/Assignment/Assignment.Tests/SampleDataTests.cs
@@ -77,15 +77,17 @@ public class SampleDataTests
     {
         // Arrange
         SampleData data = new() { CsvRows = hardCodedCSVList };
+        IPerson fremont = data.People.First();
 
         // Assert
         Assert.AreEqual<int>(6, data.People.Count());
-        Assert.AreEqual<string>("Fremont", data.People.First().FirstName);
-        Assert.AreEqual<string>("Pallaske", data.People.First().LastName);
-        Assert.AreEqual<string>("fpallaske3@state.gov", data.People.First().EmailAddress);
-        Assert.AreEqual<string>("AL", data.People.First().Address.State);
-        Assert.AreEqual<string>("Atlanta", data.People.First().Address.City);
-        Assert.AreEqual<string>("16958 Forster Crossing", data.People.First().Address.StreetAddress);
+        Assert.AreEqual<string>("Fremont", fremont.FirstName);
+        Assert.AreEqual<string>("Pallaske", fremont.LastName);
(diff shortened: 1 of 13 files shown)


Assignment/Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,119 @@
...
+
+    [TestMethod]
+    public void FilterByEmailAdreess_GivenFilter_ReturnsFiltered()
+    {
+        // Arrange
+        SampleData data = new() { CsvRows = hardCodedCSVList };
+        List<string> expected = new() { "Fremont Pallaske", "Priscilla Jenyns" };
+
+        // Act
+        IEnumerable<string> actual = data.FilterByEmailAddress(EmailFilter).Select(person => $"{person.FirstName} {person.LastName}");
```
VeraKorchemnaya, line 99: EmailFilter can be converted to lambda expression instead

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,58 @@
...
+                Address address = new(elements[4], elements[5], elements[6], elements[7]);
+                return new Person(elements[1], elements[2], address, elements[3]);
+            }
 
-        // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+            return CsvRows.Select(MakePersons)
+                .OrderBy(person => person.Address.State)
+                .ThenBy(person => person.Address.City)
+                .ThenBy(person => person.Address.StreetAddress);
```
VeraKorchemnaya, line 42: Need to change this to zip and not StreetAddress

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,57 @@
...
+    // 1.
+    public IEnumerable<string> CsvRows { get => _CsvRows; init { _CsvRows = value; } }
+    private IEnumerable<string> _CsvRows = File.ReadAllLines("People.csv").Skip(1);
+
+    // 2.
+    public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
     {
-        // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        return CsvRows.Select(row => row.Split(",")[6]).Distinct().OrderBy(state => state);
```
mmwoodfo, line 17: You could consider using a private helper method that parses the row into a person object, that way you don't need to use the indexs everywhere and can instead use dot operators.

e.g. you'd call a private method called `ParseRow` or `ParsePerson` here that would parse the row into a person object and return it, then you could just do ReturnedPerson.State instead of indexing into location `6`

mmwoodfo (commented):
Looks great. Good job!

PR #175, opened 2023-03-10
Assignment7+8 linq

gknutson1:


Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,46 @@
...
 
-namespace Assignment
+public class SampleData : ISampleData
 {
-    public class SampleData : ISampleData
-    {
-        // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+    // 1.
+    public IEnumerable<string> CsvRows => File.ReadLines(@"People.csv").Skip(1);
```
mmwoodfo, line 6:
```suggestion
    public IEnumerable<string> CsvRows => File.ReadLines("People.csv").Skip(1);

```
You shouldn't need `@` here

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,46 @@
...
-        // 5.
-        public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(
-            Predicate<string> filter) => throw new NotImplementedException();
+    // 5.
+    public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(Predicate<string> filter)
+    {
+        
+        string[] mails = Array.FindAll<string>(People.ToArray().Select(x => x.EmailAddress).ToArray(), filter);
+        List<(string, string)> items = new List<(string, string)>();
+        foreach (Person i in People)
```
mmwoodfo, line 25: You should be using linq's select here, rather than foreach

PR #179, opened 2023-03-17
Assignment9+10 multithreading - Hunter Thompson & Hunter Allen

StrangeRanger:
@hallen10


Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -109,12 +135,13 @@ async public Task RunAsync_MultipleHostAddresses_True()
     }
```
slanante, line 135: This test method fails:
image
You can fix this by having stringBuilder append to PingResult's StdOutput somewhere within this method:
image
In this screenshot you can see that stringBuilder is null within this method at the return statement:
image

slanante (commented):
Great job! You guys got most of the requirements met, I pointed out a possible issue and fix below. Also, I think you may want to remove these as it will allow you to see possible errors:
image

Implement PingProcess' public Task RunTaskAsync(string hostNameOrAddress) ✔
Implement PingProcess' async public Task RunAsync(string hostNameOrAddress) ✔
Complete/fix AND test async public Task RunAsync(IEnumerable hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
Implement AND test public Task RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task is also okay. NOTE: This method does NOT use Task.Run.✔

lsobtafo:
Implement PingProcess' public Task RunTaskAsync(string hostNameOrAddress) ✔
Implement PingProcess' async public Task RunAsync(string hostNameOrAddress) ✔
Complete/fix AND test async public Task RunAsync(IEnumerable hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
Implement AND test public Task RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task is also okay. NOTE: This method does NOT use Task.Run.✔

Good job for the most part!

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -109,12 +135,13 @@ async public Task RunAsync_MultipleHostAddresses_True()
     }
 
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
-    async public Task RunLongRunningAsync_UsingTpl_Success()
+//#pragma warning disable CS1998 // Remove this
```
mmwoodfo, line 138: Remove this

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,24 +27,59 @@ public PingResult Run(string hostNameOrAddress)
 
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        StartInfo.Arguments = hostNameOrAddress;
```
mmwoodfo, line 30: Notice how this method has the same body as RunAsync, you can simply return the result of RunAsync below.

e.g. 
` return RunAsync(hostNameOrAddress);`

PR #180, opened 2023-03-17
Assignment (9 + 10) @ lsobtafo

slanante:
Initial Pull Request, will add more content over the weekend.

### DIFF
--- a/Assignment/Assignment/PingProcess.cs
+++ b/Assignment/Assignment/PingProcess.cs
@@ -27,7 +27,7 @@ public class PingProcess
 
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        return Task.Run(() => Run(hostNameOrAddress));
     }
 
     async public Task<PingResult> RunAsync(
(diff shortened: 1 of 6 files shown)


slanante (commented):
Great job! You guys got most of the requirements met, I pointed out a possible issue and fix below. Also, I think you may want to remove these as it will allow you to see possible errors: 
image


Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.✔

Assignment/Assignment/PingProcess.cs
```diff
@@ -20,47 +20,71 @@ public PingResult Run(string hostNameOrAddress)
...
-        string hostNameOrAddress, CancellationToken cancellationToken = default)
+        IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        cancellationToken.ThrowIfCancellationRequested();
+
+        StringBuilder? stringBuilder = null;
+        ParallelQuery<Task<int>> all = hostNameOrAddresses.AsParallel().Select(async item =>
```
mmwoodfo, line 39: This one shouldn't be using a parallel query, as it should only take a single host name, as opposed to the one bleow, which takes an array

PR #181, opened 2023-03-19
Assignment9+10 multithreading

j0shh3ss:
Bhopss152
Brendan Hopkins

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -107,7 +107,7 @@ public class PingProcessTests
     {
         var pingProcess = new PingProcess();
         var cancellationTokenSource = new CancellationTokenSource();
-        cancellationTokenSource.CancelAfter(100); // cancel after 100 ms
+        cancellationTokenSource.CancelAfter(100);
         try
         {
             await pingProcess.RunAsync("google.com", cancellationTokenSource.Token);
@@ -130,7 +130,7 @@ public class PingProcessTests
     {
         var pingProcess = new PingProcess();
         var cancellationTokenSource = new CancellationTokenSource();
-        cancellationTokenSource.CancelAfter(100); // cancel after 100 ms
+        cancellationTokenSource.CancelAfter(100);
         try
         {
             await pingProcess.RunAsync("google.com", cancellationTokenSource.Token);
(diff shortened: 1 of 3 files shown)


AnnikaStachowiak (commented):
Instructions

        StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
    Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.

Fundamentals

    Place all shared project items into a Directory.Build.targets file.❌
    For this assignment, consider using Assert.AreEqual<T>() (the generic version) ✔

RungLongRunningAsync test and StringBuilderAppendLine test do not pass. Langversion is not set. Just a couple of small issues, but looks good otherwise.

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -53,79 +54,157 @@ public void Run_CaptureStdOutput_Success()
...
+   // var pingProcess = new PingProcess("google.com");
 
+    // Act
+    //var pingResult = await pingProcess.RunLongRunningAsync_UsingTpl();
+
+    // Assert
+    //AssertValidPingOutput(pingResult);
+//}
+
+    private StringBuilder _stringBuilder = new StringBuilder();
```
mmwoodfo, line 185: image

Assignment/Assignment/PingProcess.cs
```diff
@@ -25,39 +25,101 @@ public PingResult Run(string hostNameOrAddress)
         return new PingResult( process.ExitCode, stringBuilder?.ToString());
     }
 
-    public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
+    public async Task<PingResult> RunTaskAsync(string hostNameOrAddress)
```
mmwoodfo, line 28: This should _NOT_ be using async/await

Assignment/Assignment/PingProcess.cs
```diff
@@ -25,39 +25,101 @@ public PingResult Run(string hostNameOrAddress)
...
     {
-        throw new NotImplementedException();
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder? stringBuilder = null;
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default);
+        var processExited = new TaskCompletionSource<bool>();
+        process.Exited += (sender, args) => processExited.TrySetResult(true);
```
mmwoodfo, line 36: You shouldn't need to deal with process.Exited here, just call RunProcessInternal and return the result of that.

Assignment/Assignment/PingProcess.cs
```diff
@@ -25,39 +25,101 @@ public PingResult Run(string hostNameOrAddress)
...
 
-    async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
+
+    async public Task<PingResult> RunAsync(string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
+        cancellationToken.ThrowIfCancellationRequested();
+
+        StartInfo.Arguments = hostNameOrAddress;
         StringBuilder? stringBuilder = null;
-        ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
```
mmwoodfo, line 44: You should keep the ParallelQuery here

PR #182, opened 2023-03-19
Assignment 9+10 multithreading - Zach Lawson and Tom Rohr

zlawson-ewu:
Assignment 9+10 Multithreading peer programmed by Zach Lawson (@zlawson-ewu) and Tom Rohr (@Tom-Rohr). No extra credit attempted.

### DIFF
--- a/Assignment/Assignment/PingProcess.cs
+++ b/Assignment/Assignment/PingProcess.cs
@@ -27,15 +27,19 @@ public class PingProcess
 
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress));
+
+        return task;
     }
 
     async public Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        while(true)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            return await RunTaskAsync(hostNameOrAddress);
+        }
     }
 
     async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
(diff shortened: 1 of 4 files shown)


Nolan-Posey (commented):
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ✔

lace all shared project properties into a Directory.Build.Props file.
and enabled .NET analyzers for both projects✔
For this assignment, consider using Assert.AreEqual<T>() (the generic version) ✔

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +28,58 @@ public PingResult Run(string hostNameOrAddress)
...
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        while(true)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            Task<PingResult> task = Task.Run(() => RunTaskAsync(hostNameOrAddress), cancellationToken);
+            return await task;
+        }
```
mmwoodfo, line 44: You don't necessarily want this to hang forever.

Something like this may be more approperate
```suggestion
        Task<PingResult> task = Task.Run(() =>
        {
            cancellationToken.ThrowIfCancellationRequested();
            Process process = RunProcessInternal( // this will simulate your long running task
                StartInfo, updateStdOutput, default, cancellationToken);
            return new PingResult(process.ExitCode, stringBuilder?.ToString());
        });
        return await task.WaitAsync(cancellationToken);
```

PR #183, opened 2023-03-20
Assignment9+10

FenrisStargazer:

I am very confused and very lost with all this. Any and all feedback would be appreciated.

### DIFF
--- a/Assignment/Assignment/PingProcess.cs
+++ b/Assignment/Assignment/PingProcess.cs
@@ -27,7 +27,8 @@ public class PingProcess
 
     public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
     {
-        throw new NotImplementedException();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress));
+        return task;
     }
 
     async public Task<PingResult> RunAsync(
--- a/Assignment/Assignment/PingProcess.cs
+++ b/Assignment/Assignment/PingProcess.cs
@@ -34,9 +34,9 @@ public class PingProcess
     async public Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress));
+        cancellationToken.ThrowIfCancellationRequested();
+        return await task;
     }
 
     async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
@@ -44,9 +44,8 @@ public class PingProcess
         StringBuilder? stringBuilder = null;
         ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
         {
-            Task<PingResult> task = null!;
-            // ...
-
+            CancellationToken cancellationToken = default;
+            Task<PingResult> task = RunAsync(item, cancellationToken);
             await task.WaitAsync(default(CancellationToken));
             return task.Result.ExitCode;
         });
@@ -56,12 +55,15 @@ public class PingProcess
         return new PingResult(total, stringBuilder?.ToString());
     }
 
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameorAddresses, CancellationToken cancellationToken = default)
+    {
+        return await RunAsync(hostNameorAddresses.ToArray(), cancellationToken);
+    }
+
     async public Task<PingResult> RunLongRunningAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        //Task.Factory.StartNew(() => RunProcessInternal());
     }
 
     private Process RunProcessInternal(
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -50,6 +50,7 @@ public class PingProcessTests
     [TestMethod]
     public void Run_CaptureStdOutput_Success()
     {
+        //TODO: figure out what's going wrong with "assertvalidpingoutput"
         PingResult result = Sut.Run("localhost");
         AssertValidPingOutput(result);
     }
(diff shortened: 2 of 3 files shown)


Bhopss152 (commented):
1. In the RunAsync_UsingTaskReturn_Success() method you declare PingResult as 'result' but after overwrite it with 'Task<PingResullts> varable also named 'result' (Compiler error)
2. In the "RunAsync_usingtp1withCancelattion_catchAgregateExepctionWrapping() method use a call to Sut.RunAsync with an argument, and then assert that 'aggregateException' is thrown
3. Some of your 'hostNameOrAddress' is written as "hostNameorAddress' missing the capital O 

First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". 
Do NOT use async/await in this implementation. ❌
Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ❌
Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await. ❌
Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ❌ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ❌ Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ❌✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ❌and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ❌ respectively.
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for and array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
The order of the items in the stdOutput is irrelevent and expected to be intermingled.
Fundamentals
For this assignment, consider using Assert.AreEqual<T>() (the generic version) ❌

Assignment/Assignment/PingProcess.cs
```diff
@@ -55,15 +60,23 @@ async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
...
+        return new PingResult(task.Result.ExitCode, progressOutput!.ToString());
     }
 
     async public Task<PingResult> RunLongRunningAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        //Task.Factory.StartNew(() => RunProcessInternal());
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder outputStrings = new();
+        return await Task.Run(() => RunLongRunningAsync(StartInfo, (line) => (outputStrings ??= new StringBuilder()).AppendLine(line), default, default));
```
gknutson1: You need to call `toString` on your StringBuilder. You can do this by recreating the PingResult using the data from Task.Run, instead of just retruning the result:
```
// The same as your code, except assign the result of task.run to a variable instead of returning it
PingResult result = await Task.Run(() => RunLongRunningAsync(StartInfo, (line) => (outputStrings ??= new StringBuilder()).AppendLine(line), default, default));

// Return a new PingResult created from the old ExitCode and the `ToString` value of your `StringBuilder`
return new PingResult(result.ExitCode, outputStrings.ToString());
```

gknutson1 (commented):
You need to fix the step 4 method by trimming the output of each ping command, and possibly making some other formatting changes as well.

Your step 5 is broken as well - `result.StdOutput` is incorrect, becuase you aren't calling `toString` on your `StringBuilder`.

I also noticed a few `Assert.AreEqual` calls that are not using generics.

4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌
   NOTE:
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay. ❌

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file. ✔
- Place all shared project items into a `Directory.Build.targets` file. ✔
- Ensure nullable reference types is enabled  ✔
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ❌

Assignment/Assignment/PingProcess.cs
```diff
@@ -55,12 +60,23 @@ async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
         return new PingResult(total, stringBuilder?.ToString());
     }
 
+    async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)
+    {
+        return await RunAsync(hostNameOrAddresses.ToArray(), cancellationToken);
+    }
```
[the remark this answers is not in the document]
FenrisStargazer: Honestly, idk. But I saw this in the teams chat regarding this assignment so I elected to attempt to follow the rather specific instructions. Even if it isn't the most concise.
image

Assignment/Assignment/PingProcess.cs
FenrisStargazer: Ah, that's what I was missing. Thanks!

Assignment/Assignment/PingProcess.cs
```diff
@@ -55,12 +59,30 @@ async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
...
-        throw new NotImplementedException();
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder outputStrings = new();
+        Action<string?>? output = (line) => {
+            lock (outputStrings ??= new())
+            {
+                outputStrings.AppendLine(line);
+            }
+        };
+        PingResult res = await Task.Run(() => RunLongRunningAsync(StartInfo, output, default, default));
```
mmwoodfo, line 84: Instead of `Task.Run` you should use `Task.Start` here

PR #189, opened 2023-03-21
Assignment9+10 multithreading  Michael Moran(@mmoran314) Ben Fristad (@BenFristadSchool)

mmoran314:


VeraKorchemnaya:
I will review.

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,55 @@ public PingResult Run(string hostNameOrAddress)
...
 
     async public Task<PingResult> RunAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
+        cancellationToken.ThrowIfCancellationRequested();
+        Task<PingResult> task = Task.Run(() => Run(hostNameOrAddress), cancellationToken);
         await task;
-        throw new NotImplementedException();
+        return task.Result;
```
VeraKorchemnaya, line 41: It doesn't make too much sense to call both await and .Result because they kind of do similar things, I would just use await like this: 

return await task;

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -59,42 +60,59 @@ public void RunTaskAsync_Success()
...
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
-        
+        CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+        CancellationToken cancellationToken = cancellationTokenSource.Token;
+        cancellationTokenSource.Cancel(); 
+        Task<PingResult> result = Sut.RunAsync("localhost",  cancellationToken);
+        AssertValidPingOutput(result.Result);
```
VeraKorchemnaya, line 95: To properly test Cancel you should probably put it after the call to Sut.RunAsync. If you do that though your test is not going to pass, so you need to modify a couple things. It also seems unnecessary to me that you call AssertVaidPingOutput here. You will also want to wait (or .Result) on the Task to make sure that the method will either complete or will throw the exception. 
```suggestion
        CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
        CancellationToken cancellationToken = cancellationTokenSource.Token;
        Task<PingResult> result = Task.Run(() => Sut.RunAsync("localhost", cancellationToken));
        cancellationTokenSource.Cancel();
        result.Wait();
```

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -59,42 +60,59 @@ public void RunTaskAsync_Success()
...
+        Task<PingResult> result = Sut.RunAsync("localhost", cancellationToken);
+        try
+        {
+            AssertValidPingOutput(result.Result);
+        }
+        catch(AggregateException ex)
+        {
+            Exception taskCanceledException = ex.Flatten();
+            throw taskCanceledException.InnerException!;
+        }
```
VeraKorchemnaya, line 115: Again here you want to be calling Cancel after the call Sut.RunAsync but if you do that your test won't pass so you'll need to fix some other things. 

```suggestion
        CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
        CancellationToken cancellationToken = cancellationTokenSource.Token;
        Task<PingResult> result = Task.Run(() => Sut.RunAsync("localhost", cancellationToken));
        cancellationTokenSource.Cancel();
        try
        {
            _ = result.Result;
        }
        catch(AggregateException ex)
        {
            Exception taskCanceledException = ex.Flatten();
            throw taskCanceledException.InnerException!;
        }
```

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,55 @@ public PingResult Run(string hostNameOrAddress)
...
             await task.WaitAsync(default(CancellationToken));
-            return task.Result.ExitCode;
+            return task.Result;
         });
 
         await Task.WhenAll(all);
-        int total = all.Aggregate(0, (total, item) => total + item.Result);
-        return new PingResult(total, stringBuilder?.ToString());
+        int total = all.Aggregate(0, (total, item) => total + item.Result.ExitCode);
+        stringBuilder.Append(all.Aggregate("", (s1, s2) => s1.Trim() + s2.Result.StdOutput));
```
VeraKorchemnaya, line 58: This would be one way to do it but I don't know if it's the most efficient solution because you copy the same output twice and put everything in order in the end. Ideally you would expect the output to be intermingled.

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,40 +27,55 @@ public PingResult Run(string hostNameOrAddress)
...
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        // Test is not passing yet
+        Process task = await Task.Factory.StartNew(() => RunProcessInternal(startInfo, progressOutput, progressError, token), token,
+            TaskCreationOptions.LongRunning, TaskScheduler.Current);
+        return new PingResult(task.ExitCode, progressOutput!.ToString());
+        
     }
```
VeraKorchemnaya, line 79: This looks super close to working!! Great job! 
Take a look at what me and my partner did [Our Solution](https://github.com/IntelliTect-Samples/EWU-CSCD371-2023-Winter/pull/194/files#diff-0967781ea1daa01ae7ef4e93f9c3f592edd0c2c1a1ded1a081f1a364f6e85db1)

VeraKorchemnaya (commented):
## Instructions

1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ❌✔
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ❌ with a `TaskCanceledException` inner exception ❌ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ❌ and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ❌ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌✔
   NOTE:
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay. ❌✔ (Almost!)

## Fundamentals

- Ensure nullable reference types is enabled ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version) ✔
- All of the above should be unit tested ❌ (some tests need fixing)

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -109,14 +127,16 @@ async public Task RunAsync_MultipleHostAddresses_True()
...
+        CancellationToken cancellationToken = cancellationTokenSource.Token;
+
+        PingResult result = await Sut.RunLongRunningAsync("ping", cancellationToken);
+        
         AssertValidPingOutput(result);
     }
-#pragma warning restore CS1998 // Remove this
 
     [TestMethod]
     public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
```
VeraKorchemnaya, line 142: I forgot to mention. You should probably update this test to be thread safe because there are times when it doesn't pass. To do that you would just put a lock around the string builder.

Assignment/Assignment/PingProcess.cs
VeraKorchemnaya: Something else I should mention is that in the test for this method you pass in "Ping" as the hostname/address which is not valid. 
I know there should be a way to do it the way that you are trying to do it (which is to return PingResult) but I tried a variety of things and it keeps returning System.Action`1[System.String]. So idk, maybe consider just reworking your method to return int instead of PingResult.

mmoran314: Ah dang, I knew I was on the right track.

Assignment/Assignment/PingProcess.cs
mmoran314: I feel like this way keeps the code simple and concise, even if less efficient. I believe the output is intermingled as well but I could be wrong.

VeraKorchemnaya: Yeah, I mean that's definitely a valid point. You can print out the output for yourself and take a look. It does combine it but it's not like mixed together if you know what I mean.

PR #190, opened 2023-03-21
Assignment9+10 - Ian Kaiserman

Afterlyte:
Paired with @eroadhouse

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Assignment.Tests;
@@ -58,43 +59,55 @@ public class PingProcessTests
     public void RunTaskAsync_Success()
     {
         // Do NOT use async/await in this test.
+        PingResult result = Sut.RunTaskAsync("localhost").Result;
         // Test Sut.RunTaskAsync("localhost");
+        AssertValidPingOutput(result);
     }
 
     [TestMethod]
     public void RunAsync_UsingTaskReturn_Success()
     {
         // Do NOT use async/await in this test.
-        PingResult result = default;
+        PingResult result = Sut.RunAsync("localhost").Result;
         // Test Sut.RunAsync("localhost");
         AssertValidPingOutput(result);
     }
 
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
     async public Task RunAsync_UsingTpl_Success()
     {
         // DO use async/await in this test.
-        PingResult result = default;
-
+        PingResult result = await Sut.RunAsync("localhost");
         // Test Sut.RunAsync("localhost");
         AssertValidPingOutput(result);
     }
-#pragma warning restore CS1998 // Remove this
 
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
-        
+        CancellationTokenSource source = new();
+        source.Cancel();
+        Task<PingResult> result = Sut.RunAsync("localhost", source.Token);
+        _ = result.Result;
     }
 
     [TestMethod]
     [ExpectedException(typeof(TaskCanceledException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException()
     {
-        // Use exception.Flatten()
+        CancellationTokenSource source = new();
+        source.Cancel();
+        Task<PingResult> result = Sut.RunAsync("localhost", source.Token);
+        try
+        {
+            _ = result.Result;
+        }
+        catch (AggregateException ex)
+        {
+            throw ex.Flatten().InnerException!;
+        }
     }
 
     [TestMethod]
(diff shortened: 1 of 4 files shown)


AndrewKorchemniy:
will review

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,65 +59,75 @@ public void Run_CaptureStdOutput_Success()
...
-#pragma warning restore CS1998 // Remove this
 
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
-        
+        CancellationTokenSource source = new();
+        source.Cancel();
```
AndrewKorchemniy, line 91: I believe the goal is to call cancel after, not before, the PingProcess.RunAsync() is called. Otherwise, it will always cancel before the process is ever running. Since CancellationTokenSource also implements IDisposable, it's a good idea to call its dispose() once you're done with the object - although, it's within a test anyway so...

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,65 +59,75 @@ public void Run_CaptureStdOutput_Success()
...
+        Task<PingResult> result = Sut.RunAsync("localhost", source.Token);
+        try
+        {
+            _ = result.Result;
+        }
+        catch (AggregateException ex)
+        {
+            throw ex.Flatten().InnerException!;
```
AndrewKorchemniy, line 109: This works, but it also drops the 
```suggestion
            exception = exception.Flatten();
            ExceptionDispatchInfo.Capture(exception.InnerException ?? exception).Throw();
```
Doing it this way doesn't drop the exception info.

Assignment/Assignment/PingProcess.cs
```diff
@@ -20,32 +20,48 @@ public PingResult Run(string hostNameOrAddress)
...
     {
-        StringBuilder? stringBuilder = null;
+        StringBuilder? stringBuilder = new();
         ParallelQuery<Task<int>>? all = hostNameOrAddresses.AsParallel().Select(async item =>
         {
-            Task<PingResult> task = null!;
-            // ...
-
+            StartInfo.Arguments = item;
+            void updateStdOutput(string? line) => (stringBuilder ??= new StringBuilder()).AppendLine(line);
```
AndrewKorchemniy, line 59: StringBuilder is not thread safe. There are several approaches you could take.
The simplest is to just not use a shared StringBuilder (although the instructions say it's "expected" for them to be intertwined), and another way is to use a lock.

Assignment/Assignment/PingProcess.cs
```diff
@@ -55,12 +71,29 @@ async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
...
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder stringBuilder = new();
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+        Task<int> longRunning = RunLongRunningAsync(StartInfo, updateStdOutput, default, default);
```
AndrewKorchemniy, line 86:
```suggestion
        Task<int> longRunning = RunLongRunningAsync(StartInfo, updateStdOutput, default, cancellationToken );

```
Make sure to use the cancellation token.

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -58,65 +59,75 @@ public void Run_CaptureStdOutput_Success()
...
     {
-        PingResult result = default;
+        PingResult result = await Sut.RunLongRunningAsync("localhost");
         // Test Sut.RunLongRunningAsync("localhost");
         AssertValidPingOutput(result);
     }
-#pragma warning restore CS1998 // Remove this
 
     [TestMethod]
     public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
```
AndrewKorchemniy, line 133: I believe this test was just give us a hint on StringBuilder not being thread safe - it sometimes does not pass.
Since it's not testing anything within PingProcess (and was probably just a hint), I think it's fine to remove.

AndrewKorchemniy (commented):
## Instructions

4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for and array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ✔
   NOTE:
      - The order of the items in the stdOutput is irrelevent and expected to be intermingled.
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.
    ✔

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌

## Fundamentals

- Ensure you enable:
  - nullable reference types is enabled ✔
  - net7 targeted ✔
  - C# 11.0 ✔

Assignment/Assignment.Tests/PingProcessTests.cs
Afterlyte: We initially tried doing the cancellation after the RunAsync call and it was running before the cancellation could even happen, do you have any suggestions on how to approach that? Because I do agree that the goal probably is to do it after.

PR #192, opened 2023-03-21
Assignment9+10: With Jessica Ho

jacobnewcomb:
@JessicaH46

### DIFF
--- a/Assignment/Assignment.Tests/Assignment.Tests.csproj
+++ b/Assignment/Assignment.Tests/Assignment.Tests.csproj
@@ -1,6 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+	<TargetFramework>net7.0</TargetFramework>
+	<ImplicitUsings>enable</ImplicitUsings>
+	<Nullable>enable</Nullable>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Assignment.Tests;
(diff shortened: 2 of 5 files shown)


Afterlyte:
I will review this

Assignment/Assignment.Tests/Assignment.Tests.csproj
```diff
@@ -1,6 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+	<TargetFramework>net7.0</TargetFramework>
```
Afterlyte, line 4: Shared Properties like TargetFramework and Nullable shouldn't be here if they're in Directory.Build.Props already

Assignment/Assignment.Tests/PingProcessTests.cs
```diff
@@ -57,44 +58,42 @@ public void Run_CaptureStdOutput_Success()
...
     }
 
     [TestMethod]
     public void RunAsync_UsingTaskReturn_Success()
     {
-        // Do NOT use async/await in this test.
-        PingResult result = default;
-        // Test Sut.RunAsync("localhost");
-        AssertValidPingOutput(result);
+        AssertValidPingOutput(Sut.RunAsync("localhost").Result);
```
Afterlyte, line 68: It's a bit more readable if you separate function calls into variables (like here, having RunAsync return as a PingResult object) so it's clear what you're trying to test with at first glance.

```suggestion
        PingResult result = Sut.RunAsync("localhost").Result);
        
        AssertValidPingOutput(result);

```

Assignment/Assignment/PingProcess.cs
```diff
@@ -27,26 +27,29 @@ public PingResult Run(string hostNameOrAddress)
...
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        while (true)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            return await Task.Run(() => RunTaskAsync(hostNameOrAddress), cancellationToken);
+        }
```
Afterlyte, line 40: I'm very confused about this while loop, and it feels very risky to just have a "while true" loop in here.
…Although all the tests concerning it passed without issue so maybe I just don't quite understand it.

Assignment/Assignment/PingProcess.cs
```diff
@@ -58,9 +61,9 @@ async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
     async public Task<PingResult> RunLongRunningAsync(
         string hostNameOrAddress, CancellationToken cancellationToken = default)
     {
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        Task<PingResult> task = Task.Factory.StartNew(() =>
+            Run(hostNameOrAddress), cancellationToken, TaskCreationOptions.LongRunning, TaskScheduler.Current);
```
Afterlyte, line 65: Requirements specified to not use the Task.Run method. (Hint: look at how the Run method is implemented to figure out how to do it without using it explicitly)

Afterlyte (commented):
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ❌
  - **Not tested**

   NOTE:
6. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.
   NOTE: This method does **NOT** use `Task.Run`. ❌
  - **Run method is used**

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌

## Fundamentals

  - **Duplicate properties in other csproj files should be removed**
- Ensure nullable reference types is enabled  ✔
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔

**Great job overall! A couple minor tweaks to make but other than that you guys did really well.**

Assignment/Assignment/PingProcess.cs
JessicaH46: Thank you! Your review was very helpful and I think we were able to fix it!

Assignment/Assignment.Tests/PingProcessTests.cs
JessicaH46: thank you for the suggestion, it does provide more readability!

